PR #8, opened 2018-08-24
Adds some internal/external presentations to L&L docs

ashfurrow:
I brought these over from the [old project board](https://github.com/artsy/meta/projects/1). I also added some internal ones to get a sense of what we talk about even when the speakers aren't external.


anandaroop:
Nice!

PR #14, opened 2018-08-27
Applies artsy/meta#44

ashfurrow:
This PR got lost in the docs migration, looks like: https://github.com/artsy/meta/pull/44. /cc @mbilokonsky


orta:
Thanks!

PR #32, opened 2018-09-03
Adds comments to the inline walkthrough

orta:
This denotes:

* Team standup are optional
* That open RFCs are already published In the dev channel
* That if you're unsure about L&Ls then you can always asking the L&L channel


orta:
@ashfurrow  ^

PR #34, opened 2018-09-05
docker & docker-compose are not in caskroom

erikdstock:
Minor, but caused an issue on setup. We installed docker & docker-compose instead using plain brew.

### DIFF
--- a/practices/platform/hokusai.md
+++ b/practices/platform/hokusai.md
@@ -8,7 +8,7 @@ Docker and Kubernetes clusters.
 ### Quickstart
 
 ```
-brew install git && brew tap caskroom/cask && brew cask install docker docker-compose
+brew install git docker docker-compose
 curl https://artsy-provisioning-public.s3.amazonaws.com/hokusai -o /usr/local/bin/hokusai && chmod +x /usr/local/bin/hokusai
 hokusai configure --kubectl-version 1.6.3 --s3-bucket artsy-citadel --s3-key k8s/config
 ```


alloy:
I think the danger of this is that those tools no longer get updated in tandem with the docker daemon, no?

I think I installed docker from a dmg, which presumably is what caskroom taps from, and when it prompts me for an update my CLI tools auto-update along with it.

Having said that, I still know very little of docker, maybe these CLI tools are very thin wrappers that call into the docker daemon and thus don’t really need to be updated in tandem?

erikdstock:
I'm not sure @alloy - in our case we just got a message that docker wasn't in the caskroom. Is it possible we should just install docker for mac?

orta:
I use Docker for Mac, personally

alloy:
I see. Yeah I use Docker for Mac as well, I thought that was what cask would install too 🤷‍♂️

izakp:
> I see. Yeah I use Docker for Mac as well, I thought that was what cask would install too

Ditto.

I just successfully reinstalled docker and docker-compose using brew cask install.  One note is that we don't need to separately install docker-compose as it comes bundled with Docker for Mac.

```
isacpetruzzi@Artsy-Isac gravity (master) $ brew cask install docker
==> Satisfying dependencies
==> Downloading https://download.docker.com/mac/stable/26764/Docker.dmg
######################################################################## 100.0%
==> Verifying checksum for Cask docker
==> Installing Cask docker
==> Moving App 'Docker.app' to '/Applications/Docker.app'.
🍺  docker was successfully installed!
isacpetruzzi@Artsy-Isac gravity (master) $ which docker
/usr/local/bin/docker
isacpetruzzi@Artsy-Isac gravity (master) $ docker --version
Docker version 18.06.1-ce, build e68fc7a
isacpetruzzi@Artsy-Isac gravity (master) $ which docker-compose
/usr/local/bin/docker-compose
isacpetruzzi@Artsy-Isac gravity (master) $ docker-compose --version
docker-compose version 1.22.0, build f46880f
```

izakp:
@erikdstock did you get Docker for Mac using plain brew install?  What versions do you have?  I am on:

- MacOS 10.12.6
- Homebrew 1.7.3
- Homebrew-Cask 1.7.3

erikdstock:
I feel like there is a miscommunication happening here- I'm referring specifically to the setup instructions in this PR, which I was following with Ashley:
```
brew install git && brew tap caskroom/cask && brew cask install docker docker-compose
=> Error: Cask 'docker-compose' is unavailable: No Cask with this name exists.
```
I believe `brew cask` is about to be deprecated anyways (the ~cask~ part)... but `docker-compose` is technically in reguarly `brew`. I don't know if they would work without docker for mac or docker machine though.

izakp:
@erikdstock i see - the problem is that the `brew install docker` command gives you the Docker binary but you need to run the Docker daemon yourself, with [docker-machine and VirtualBox](https://stackoverflow.com/questions/40523307/brew-install-docker-does-not-include-docker-engine) - @bhoggard and I were using this method in the past and it was painful.

`brew cask install docker` gives you the "Docker for Mac" app, integrates with MacOS quite painlessly (has a UI and a taskbar icon for the daemon) and is what we are generally recommending.  

I'd like to keep these quickstart instructions to use brew cask for this reason, and otherwise instruct folks to download the .dmg and install Docker for Mac themselves.

Will open another PR.

izakp:
Ach accidentally just committed via the UI editing window rather than opening a PR - but here it is https://github.com/artsy/README/commit/633a198af192fbc7c508121500325c096a2145d8

izakp:
cc @ashleyjelks - would recommend to uninstall the `brew install`ed `docker` / `docker-compose` and use `brew cask install docker` or [download the Docker for Mac .dmg](https://docs.docker.com/docker-for-mac/), which contains both, bundled in an app.  Sorry for the confusion!

PR #36, opened 2018-09-07
Refreshes standup docs.

ashfurrow:
This does a few things:

- Adds a note for moving laptops away from the Zoom speakerphone.
- Removes the announcement for writing office hours, since it robs me of an opportunity to really sell the event. It made sense to include in the template when I was running standup each week, but not anymore.
- Corrects the Show & Tell time.


jonallured:
Nice one!! ❤️

PR #40, opened 2018-09-11
Adds a doc on how we find & evaluate juniors

orta:
A lot of the discussion around this doc happened in private, but now it's time for it to get some sunlight 👍 

This PR also updates some of our TOC around the docs, which happened before I got CI running.


orta:
I think this'll be foundations for when that role exists 👍

orta:
merge on green 👍

PR #47, opened 2018-09-17
Minor updates to clarify expectations around on-call

sweir27:
This PR adds two things to the support doc:
- A note about managers covering
- A short list of expectations (tries to make explicit what is explained in greater detail elsewhere)

### DIFF
--- a/playbooks/support.md
+++ b/playbooks/support.md
@@ -22,7 +22,7 @@ which is determined at least a month in advance (giving people ample time to mak
 
 The schedule is configured on the
 [Engineering On-Call calendar](https://calendar.google.com/calendar/embed?src=artsymail.com_nolej2muchgbpne9etkf7qfet8%40group.calendar.google.com&ctz=America%2FNew_York).
-Trading shifts (because of vacations, obligations, etc.) is encouraged as long as the calendar is kept up-to-date.
+Trading shifts (because of vacations, obligations, etc.) is encouraged as long as the calendar is kept up-to-date. Please address any scheduling issues as early as possible.
 
 During work hours, on-call engineers are responsible for responding to issues in
 [the #incidents slack channel](https://artsy.slack.com/messages/C9RK0BLEP/) in a timely fashion. Outside of work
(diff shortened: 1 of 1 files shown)


orta:
👍

playbooks/support.md
```diff
@@ -31,6 +31,9 @@ hours, engineers are _only_ responsible for downtime issues.
 The two engineers who are on-call are free to split the responsibility for monitoring outside of work hours, but we
 suggest either doing every other day or first half/second half.
 
+### A note about team members who leave
+If someone who is scheduled to be on-call ends up leaving the engineering team (by moving to a different team or leaving Artsy altogether), it is their (previous) manager's responsibility to cover their shift. The manager will either take the shift themselves or, if they are unavailable, find someone who can.
```
joeyAghion: This looks accurate but is a special case so should maybe fall below the important stuff. You could add a "special cases" section at the end and maybe mention the holiday rotation schedule as well.

playbooks/support.md
```diff
@@ -143,6 +146,21 @@ You can find the point people in the [Potential Wiki](https://github.com/artsy/p
...
+- Address all scheduling issues/conflicts for your shift as early as you know about them.
+- Communicate to your manager when something comes up that prevents you from being on-call.
+- Lean on the entire team for help-- no one expects you to know how to debug every system.
+- Improve the process where you can (including educating others).
+- Help facilitate solutions, if need-be.
+
+### What is not expected of you
+- To know the ins and outs of every system.
+- To wake up at night for a non-emergency.
+- To fix all of the bugs you find.
```
joeyAghion: This is good stuff but is basically the short version of the longer list above. Ironically, this makes the doc _longer_ and harder to follow, because now there are 2 separate lists of what to do. Would be even better to integrate these above (maybe in bold for clarity) and shorten the rest.

alloy: I think that in a case like this, where people may be anxious about their responsibilities, a bit longer and explicit in what they do _not_ need to worry about can be helpful.

sweir27:
I attempted to both emphasize and consolidate the information I'd added-- mind taking another look @joeyAghion ?

PR #52, opened 2018-09-21
@orta => Add a section about testing

eessex:
Wanted to get the ball rolling on this section as I've recently noticed some new apps without tests.


orta:
People are writing apps without tests !?!

Yeah, this is definitely good

PR #53, opened 2018-09-24
Adds front-end office hours to the announcements

orta:
This notes that we're doing it


alloy:
We’re going legit 👍 

Merge on green.

PR #58, opened 2018-10-01
Update support.md

orta:
We were wondering what the right way to address timely non-incidents, and felt like talking to the PM would be the right way to raise awareness.

( I changed the order of the examples so that it would flow better )

### DIFF
--- a/playbooks/support.md
+++ b/playbooks/support.md
@@ -36,7 +36,7 @@ suggest either doing every other day or first half/second half.
 While on-call, you are accountable for investigating and fixing timely issues, escalating to additional
 point-people and/or routing to team-specific backlogs where appropriate.
 
-Your top priority during an on-call period is to address critical issues. Use your judgment about participating in
+Your top priority during an on-call period is to address critical issues. Use your judgement about participating in
 ongoing sprint and team activities.
 
 #### 1. Be available to answer requests and respond to immediate issues
--- a/playbooks/support.md
+++ b/playbooks/support.md
@@ -168,14 +168,23 @@ with. People reporting problems are still learning the process too, so a respons
 This can be a lot more effective, and more in-line with Artsy's values, than "What you've reported isn't an
 incident, talk to your PM." Remember, we're all on the same team!
 
+### Incidents
+
+- Multiple simultaneous Intercom support requests from partners (payments not succeeding, login failures)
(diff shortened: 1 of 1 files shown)


playbooks/support.md
```diff
@@ -36,7 +36,7 @@ suggest either doing every other day or first half/second half.
 While on-call, you are accountable for investigating and fixing timely issues, escalating to additional
 point-people and/or routing to team-specific backlogs where appropriate.
 
-Your top priority during an on-call period is to address critical issues. Use your judgment about participating in
+Your top priority during an on-call period is to address critical issues. Use your judgement about participating in
```
sweir27: 🤔 I think `judgement` is more of a british english thing haha

playbooks/support.md
```diff
@@ -168,14 +168,23 @@ with. People reporting problems are still learning the process too, so a respons
...
+- _Any_ automated alert that shows up in #incidents
+
 ### Non-incidents
 
 - Charge details for partners need updating
 - Single partner cannot access CMS
 - Artwork appearance on site ("How come it's not showing up in...")
 
-### Incidents
+### Urgent Non-incident
```
sweir27: This should be plural to match above

playbooks/support.md
```diff
@@ -168,14 +168,23 @@ with. People reporting problems are still learning the process too, so a respons
...
 
-- Multiple simultaneous Intercom support requests from partners (payments not succeeding, login failures)
-- Auction artworks are not available before auction open
-- _Any_ automated alert that shows up in #incidents
+- An auction bid which needs changing before an auction
+- [...]
+
+For issues like this it's best to contact the related PM, to ask for their advice. This is so that we can
```
sweir27: It can be a little tricky to segment issues in this way, and I'd almost rather us delegate anything which might fall under "urgent non-incident" into treating it as an actual "incident" or having a product solution.

Things like auction bids needing changing and subscription issues are _known_ issues which require developer intervention. For these, we should be diligent about alerting the PM and delegating the fix/feature to a relevant product team (so, if it happens again it's the responsibility of the responsible team and not the on-call folks). This frees up those on-call to deal with `incidents`.

orta:
I think this one is probably worth canning for now, if it comes up again we can re-think

PR #60, opened 2018-10-03
Rename recommendations --> resources

sweir27:
This renames the "recommendations" folder to "resources" because the folder is full of links for how to learn about various aspects at artsy (which seems much more in line with "resources").

Happy to change back if there are objections.

It also adds a link to the `what-happens-when` github repo which talks about (in short), the internet.


orta:
Works for me - change is change

PR #62, opened 2018-10-05
k8s upgrade

izakp:
Upgrade to Kubernetes 1.10.7


izakp:
Merging this as the upgrade is out and folks need to run the updated command

PR #68, opened 2018-10-12
Reorganizes support section and adds a doc for how to schedule an on-call round

sweir27:
Since you just reorganized I'm curious if you think `support` still belongs under `playbooks`. I expanded the nested folder because the docs in there may start to grow.

The meat of this change is a new doc which describes how I scheduled the latest round of on-call. I'm incredibly eager for improvements to the process-- feel free to identify those here, or just evaluate how it's described.


joeyAghion:
2 mild preferences of mine:
* avoid burying important docs like the main support playbook too deeply
* avoid anemic READMEs

One slight reorganization you could do is to put the main how-to-support content into `support/README.md` (with links to "advanced" resources like how to schedule on-call shifts, as necessary). This way it's only one click from the playbooks folder at any time in Github's UI. 

Another option would be to leave `support.md` in the top-level and put only advanced content in the subfolder. (Because this puts 2 `support...` items in the playbooks directory and probably necessitates an anemic README in the support folder, I think I prefer the other option.)

Also, note that there may be a bunch of links to the support doc from outside this repo (e.g., slack channel descriptions) that should be updated.

playbooks/support/scheduling.md
```diff
@@ -0,0 +1,31 @@
...
+  - Invite the on-call person
+  - Name the event "<their name> On-Call"
+  - Check the box so the guest can modify the event
+  - Make sure to say "yes" for sending email invites
+  - I usually create one this way and then duplicate it so all of the settings remain the same
+8. Double-check that the calendar invites look okay, then announce in slack that scheduling has finished. Something like:
+```
+Hi team! Just finished creating calendar invites for this round of on-call. If something looks wrong, please let me know! Soon I will be deleting the Notion doc so we can use this as the source of truth. Note that you can still switch shifts if something comes up!
+```
+9. After a couple of days, delete the Notion doc to avoid confusion.
```
joeyAghion: 10. Schedule a reminder to repeat this process for the next round. 😄

sweir27:
Agree about anemic READMEs, although I also don't want a bloated support doc either. /shrug

alloy:
> Also, note that there may be a bunch of links to the support doc from outside this repo (e.g., slack channel descriptions) that should be updated.

FWIW There’s already an example in the repo of a doc that still exists but only points to the new one (not sure offhand which one it is, though)

playbooks/README.md
```diff
@@ -31,7 +31,7 @@ practice should be incorporated, submit a PR!
 | [How to run a retrospective](/playbooks/retrospectives.md#readme) | The why and how for running a retrospective |
 | [How to create an RFC at Artsy](/playbooks/rfcs.md#readme) | The steps needed to request cultural changes |
 | [How to run a Lunch & Learn](/playbooks/running-lunch-and-learn.md#readme) | How to handle the people process for Lunch & Learns |
-| [The Artsy Engineering Support process](/playbooks/support.md#readme) | A process overview and some tips and tricks. |
+| [The Artsy Engineering Support process](/playbooks/support/support.md#readme) | A process overview and some tips and tricks. |
```
alloy: I wasn’t able to find this fragment (or MD header) in that doc. Is it meant to be `#engineering-support--artsy`? (In which case, maybe just leave it off entirely?)

Issue #23, opened 2018-08-27
Add instructions on how to take a "tour of Artsy"


sweir27:
cc @xtina-starr for the idea!

One of the items in our onboarding talks about taking some time to tour around Artsy.net in a staging environment, doing things like bidding in an auction and making an inquiry.

It would be cool if we had a doc that outlined some of these and provided some high-level instructions on how to actually do them. 😄

dleve123:
Closed by https://github.com/artsy/README/pull/70

PR #70, opened 2018-10-12
Adds "Tour of Artsy" onboarding session doc

sweir27:
You may also object to the anemic readme here, but I think we should start collecting documentation for the onboarding sessions we run (that way _anyone_ can run them, and there's a concrete reference).

This adds a doc for the "Artsy from the Collector's side", which we've been referring to as the "tour of artsy". It'll close this issue: https://github.com/artsy/README/issues/23

cc @dleve123 since you just went through this with me!

### DIFF
--- a/onboarding/sessions/tour-of-artsy.md
+++ b/onboarding/sessions/tour-of-artsy.md
@@ -1,0 +1,162 @@
+# Tour of Artsy
+
+This session will take you through some of our key features on artsy.net and show you how to act as a collector in our **staging environment**.
+
+There are also links to relevant code and admin interfaces where applicable.
+
+## What can you do as a user on Artsy?
+- Look around/read
+  - Browse the site, read articles on our magazine
+- Casually participate
+  - Follow artists/artworks/categories, register for auctions
+- Buy things
+  - Through auctions, e-commerce, inquiries
+
+#### Caveats about staging environment
+- Data gets copied from production —> staging once a week (so, will be a week old + data created in staging during the week gets wiped)
+- We only copy gravity data, meaning there can be some issues when referencing other services in staging (i.e. auctions)
+- Elasticsearch indices are often out-of-date
+
+#### Set-up needed before this session
+1. Find a work by a test partner that can be inquired on
+2. Create a mock auction that is online-only (and update it so it ends far enough in the future)
+3. Create a mock auction that is live
+4. Create a mock auction that requires bidder approval
+5. Find a work that can be bought via ecommerce
+
+_Tips_:
+  - Use the test card 4242 42424 4242 4242 for all transactions
+
+### Things you can do as a user just browsing around...
+1. Sign up for an account/log in (have them sign up with an artsymail account, like sarah+testing@artsymail.com)
+  - Sign up modals reflect recent work to consolidate auth modals/behavior
+  - Reference:
+    - [User model](https://github.com/artsy/gravity/blob/master/app/models/domain/user.rb)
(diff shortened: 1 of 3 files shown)


onboarding/sessions/tour-of-artsy.md
```diff
@@ -0,0 +1,162 @@
...
+    - Reference:
+      - [Artist model](https://github.com/artsy/gravity/blob/master/app/models/domain/artist.rb)
+      - [Artist app code (Force)](https://github.com/artsy/force/tree/master/src/desktop/apps/artist)
+      - [Artist app code (Reaction)](https://github.com/artsy/reaction/tree/master/src/Apps/Artist)
+  - **Auction page**
+    - Reference:
+      - [filter/sale_artworks API](https://github.com/artsy/gravity/blob/master/app/api/v1/filter_endpoint.rb)
+      - [Auction app code](https://github.com/artsy/force/tree/master/src/desktop/apps/auction)
+  - **Gallery profile page**
+    - Vanity URL since it may be shared
```
dleve123: minor, suggested rewording: "URL structure isn't namespaced by the Gallery resource for SEO/marketing reasons. Galleries are accessed by vanity URLs, for example: https://staging.artsy.net/pace-slash-macgill-gallery"

onboarding/sessions/tour-of-artsy.md
```diff
@@ -0,0 +1,162 @@
...
+    - Reference:
+      - [App code](https://github.com/artsy/force/tree/master/src/desktop/apps/partner)
+  - **Article page**
+    - Many different types of articles, determines formatting/content options
```
dleve123: not sure what you mean here: Are there different article types, and can each type have differing format options?

onboarding/sessions/tour-of-artsy.md
```diff
@@ -0,0 +1,162 @@
...
+    - Reference:
+      - [App code](https://github.com/artsy/force/tree/master/src/desktop/apps/article)
+  - **Fair microsite**
```
dleve123: Is https://github.com/artsy/vanity the repo that backs all of these vanity URLs? If so, thoughts on including this as a reference to here, and maybe the `*Gallery profile page**` page above?

onboarding/sessions/tour-of-artsy.md
```diff
@@ -0,0 +1,162 @@
...
+6. Save an artwork for later
+  - Reference:
+    - [CollectedArtwork model](https://github.com/artsy/gravity/blob/master/app/models/domain/collected_artwork.rb)
+
+### Ways that you can buy art as an Artsy user...
+1. Inquire on an artwork that is owned by a known test account (i.e. Invoicing Demo Partner)
+  - Inquiries are sent to gallery contacts that are subscribed to receive inquiry messages
+  - Galleries can respond via Conversations (in CMS) or email
+  - Collectors can respond via Messaging (in the iOS app) or email
+  - Inquiries are a key metric for our teams as it can indicate an intent to buy (and eventually lead to a purchase)
```
dleve123: I think one interesting fact for new team members joining Artsy is that purchases have, historically, not been a first-class tracked thing in the Artsy domain model. While, I think, this is changing with the BNMO effort, I would recommend being explicit with this in these docs.

dleve123: Well, not _all_ purchases are tracked in Artsy. Invoices due map to purchases, IIRC, but I believe Chung-Yi mentioned in his onboarding session that many purchases happen out of Artsy.

PR #74, opened 2018-10-15
Add poke on adding Slack avatar pic

jonallured:
This is not so important, but I think it makes one feel more like they are a part of the team. Plus no one likes looking at those default Slack avatars! 😝


damassi:
@jonallured - can you also update with a mention about JIRA? Also helpful

orta:
Yes!

jonallured:
Thanks friends, going to merge this.

PR #75, opened 2018-10-15
Clarify who runs standup

jonallured:
This was made slightly more complicated by the staggered starts to on-call shifts, so I'm just clarifying things here.


starsirius:
😍

sweir27:
🙏 thank you for adding this!!

PR #80, opened 2018-10-16
Update Hokusai's README to use correct version of kubectl

dleve123:
* s/rails/node in node example
* Add pointer to set AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID in the
  quickstart section


izakp:
Thanks for the updates!

PR #81, opened 2018-10-17
Moves docs-updating tip to top of list

ashfurrow:
Talked this over with @pepopowitz at writing office hours and we think it makes more sense to put this upfront, rather than as a pseudo-actionable item near the end.


ashfurrow:
Ashkan is out of office, going to assign to @jonallured instead.

jonallured:
Sorry for the delay, this looks great, merging!! ❤️

ashkan18:
🙌

PR #83, opened 2018-10-18
Adds useful L&L videos

sepans:
These are the ones I personally found useful. Looking for suggestions 🙌 

Wanted to add tech-stack videos but didn't know which one (if any) is the most relevant.


sepans:
Added 🔒to private links thanks to @dblandin. Merging for now.

PR #85, opened 2018-10-18
@sweir27: List work-relevant slack channels in slack doc

joeyAghion:
The `slack.md` doc lists a zillion channels _not_ relevant to work, but not the most relevant ones.


orta:
For background on this, I intentionally didn't add these dev slacks because they change so often - but I don't mind having too much potentially stale slack channels as this isn't 100% definitive of all dev slack etc

dblandin:
What do you think of adding a mention of the [#onboarding-eng](https://artsy.slack.com/messages/CCC37HXE0) 🔒Slack channel for new team members?

That has been a great resource for me this week during on-boarding.

PR #99, opened 2018-11-02
Adds initial podcast docs

ashfurrow:
These are just a first stab. We tried to record an episode but got caught up in technical problems, so we documented the solutions to those problems.

There's some noise in the PR too, not sure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ We split this repo out into different sections, each one has a summary of what's
 | [Events at Artsy](/events#readme) | Documentation on regularly occurring events and meetings. |
 | [Onboarding Notes for New Engineers](/onboarding#readme) | Your first steps to being productive |
 | [Playbooks](/playbooks#readme) | Tips, procedures, and best practices |
+| [Artsy Engineering Podcast](/podcast#readme) | How we (want to) record an engineering podcast |
 | [Engineering Practices](/practices#readme) | How do we handle cross-functional concerns. |
 | [Engineering Recommendations](/resources#readme) | Collections of further reading. |
 <!-- end_toc -->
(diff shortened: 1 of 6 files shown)


dleve123:
@ashfurrow - just curious what you think the next step should be for this PR. Wondering if it should be closed considering it’s pretty old? Also seems like there haven’t been a lot of conversations going on recently about podcasting so maybe it’s not super relevant (very possible that I’m just not part of those conversations though)?

ashfurrow:
@dleve123 thanks for the bump – just confirmed that this is still a WIP, so let's leave this PR open for now 👍

eessex:
Out of curiosity, what technical problems did you run into recording the first episode?

ashfurrow:
@eessex If I recall, we had issues recording two microphones at once in GarageBand. Actually, I did some [home recording](https://soundcloud.com/ash-furrow/fallow) and got to use some of what I learned in GarageBand there to record vocals/guitar at the same time.

eessex:
Happy to help with this process and can also recommend some alternative software to garage band for recording -- just learned all these skills to put together a radio show I'm working on, so the tools are pretty fresh in my mind.

dblandin:
> Happy to help with this process and can also recommend some alternative software to garage band for recording -- just learned all these skills to put together a radio show I'm working on, so the tools are pretty fresh in my mind.

Nice :100:

I'm interested in what your setup looks like. @zephraph might also have thoughts after putting together a pretty sweet home setup. Do we already have mics?

SamRozen:
@ashfurrow @pepopowitz I was looking at the open PRs and readme and this one has been hanging for a while? What do you think? Should we close it?

pepopowitz:
I agree that this is out of date and should be closed. I'm curious if @ashfurrow thinks any of it should be preserved, but it doesn't match how we've been working on this iteration of the podcast.

ashfurrow:
Nope, we can close this 👍

Issue #2, opened 2018-08-21
Create a doc which states what shouldn't be Open


orta:
If OSS by Default means starting with open, what are the right arguments for saying something should be closed/private.

Re: https://github.com/artsy/potential/issues/166#issuecomment-412677219

joeyAghion:
Since I brought this question up in another forum, here are a few proposals:

Always private:
* contact or role details for individuals
* any information about candidates or new hires before they've consented to announce they're move (this would include onboarding issues)
* production data or logs (including indirectly through migration output, etc.)
* repositories representing substantial business value, novelty, or criticality (such as the main API, content management system, or auction operator tools)

Maybe should be private, depending on sensitivity:
* security or compliance topics, including pending security work items or GDPR procedures
* incident/support instructions, including data recovery or incident-mitigation procedures
* sensitive business rules, definitions, or implementations (tiering, targeting, buyer/bidder qualification, spam heuristics, search ranking algorithms, data partnerships...)
* un-announced feature work
* industry or user research or analysis that may hold some competitive value
* references to private content (according to the criteria above) including documentation of private systems, links to private gists that may include private data or logs, etc.

The "maybe" category is pretty large and nebulous. Any more concrete criteria we can introduce there could be helpful. Especially valuable would be ways to _enforce_ these considerations over commits or discussions, since lots of content that is already public overlaps with these potentially-sensitive topics.

starsirius:
I'm struggling with the same single-source-of-truth vs. open by default, and want to provide one example recently while working on an onboarding session for new hires.

The [tour of collector experience](https://github.com/artsy/README/blob/345c6c0a376b9854e306ca53e8da7ec96e3ff355/onboarding/sessions/tour-of-collector-ux.md) is a great session to show case Artsy experience as a collector for new hires and is perfect to be public. However, when working on the counterpart, tour of Artsy partner experience, I realized it may not be a good candidate to be public. it includes many internal processes and admin workflow. It's not ideal I have to put our onboarding sessions in different places.

orta:
Yeah, I agree with all of the always and maybes there @joeyAghion 

I don't think there is a one-size fits answer all unless we move all documentation private, and personally I'd rather that didn't happen - when we had them private . Yet, if folks feel strongly about it though, we can have an RFC for it though!

[...]

PR #106, opened 2018-11-09
Try to document where documents should go

joeyAghion:
It resurfaced today that there's continuing ambiguity about where different types of documents should live. This is an initial stab at describing what might be reasonable "defaults," given our current family of forums and repositories. It also copies some of the discussion about public vs. private content from https://github.com/artsy/README/issues/2.

Closes https://github.com/artsy/README/issues/105.

### DIFF
--- a/playbooks/README.md
+++ b/playbooks/README.md
@@ -23,6 +23,7 @@ practice should be incorporated, submit a PR!
 | [Conferences](/playbooks/conferences.md#readme) | Attending or speaking at a conference |
 | [Data Migrations](/playbooks/data-migrations.md#readme) | Preparing, reviewing and executing data migrations |
 | [Deployments](/playbooks/deployments.md#readme) | How systems are deployed at Artsy |
+| [Documentation](/playbooks/documentation.md#readme) | How and where to document or share content |
 | [Engineer workflow](/playbooks/engineer-workflow.md#readme) | How we work together |
 | [Extracting Services](/playbooks/extracting-services.md#readme) | When and why to consider extracting or decoupling systems |
 | [GraphQL Schema Design](/playbooks/graphql-schema-design.md#readme) | What are our best practices for GraphQL Schema Design? |
@@ -33,7 +34,6 @@ practice should be incorporated, submit a PR!
 | [How to run a retrospective](/playbooks/retrospectives.md#readme) | The why and how for running a retrospective |
 | [How to create an RFC at Artsy](/playbooks/rfcs.md#readme) | The steps needed to request cultural changes |
 | [How to run a Lunch & Learn](/playbooks/running-lunch-and-learn.md#readme) | How to handle the people process for Lunch & Learns |
-| [The Artsy Engineering Support process](/playbooks/support#readme) | A process overview and some tips and tricks. |
 | [Technology choices](/playbooks/technology-choices.md#readme) | Evaluating and adopting new technology at Artsy |
 <!-- end_toc -->
 <!-- prettier-ignore-end -->
--- a/playbooks/documentation.md
+++ b/playbooks/documentation.md
@@ -19,12 +19,11 @@ E.g.:
 - Definitions of domain models
 - API documentation
 
-It _almost_ goes without saying, but other repo-specific interactions such as presenting the rationale for a design
-or offering code-review should leverage the built-in mechanisms of
-[commit messages](/playbooks/engineer-workflow.md#commits),
+Other repo-specific interactions such as presenting the rationale for a design or offering code-review should
+leverage the built-in mechanisms of [commit messages](/playbooks/engineer-workflow.md#commits),
 [pull requests](/playbooks/engineer-workflow.md#pull-requests), and comments.
 
-## Public versus private
+## Public versus private content
 
 When creating a repo, contributing a doc, or discussing in-progress work, consider whether it's appropriate for
 this content to be public or private.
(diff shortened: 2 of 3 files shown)


playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+
+- onboarding resources
+- team procedures and playbooks
+
+The private [artsy/potential](https://github.com/artsy/potential) repo supplements that with internal resources
+such as:
+
+- documentation and diagrams of internal architecture
+- security bounty program procedures
+- playbooks for handling common on-call incidents (see [the wiki](https://github.com/artsy/potential/wiki))
```
joeyAghion: The potential wiki is a weird exception that I'd be open to migrating into the repo (as discussed).

playbooks/README.md
```diff
@@ -33,7 +34,6 @@ practice should be incorporated, submit a PR!
 | [How to run a retrospective](/playbooks/retrospectives.md#readme) | The why and how for running a retrospective |
 | [How to create an RFC at Artsy](/playbooks/rfcs.md#readme) | The steps needed to request cultural changes |
 | [How to run a Lunch & Learn](/playbooks/running-lunch-and-learn.md#readme) | How to handle the people process for Lunch & Learns |
-| [The Artsy Engineering Support process](/playbooks/support#readme) | A process overview and some tips and tricks. |
```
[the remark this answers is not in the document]
orta: It should, yes

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+  management system, or auction operator tools)
+
+Content that _maybe_ should be private, depending on its sensitivity:
+
+- Security or compliance topics, including pending security work items or vulnerability-handling procedures
+- Incident/support instructions, including data recovery or incident-mitigation procedures
+- Sensitive business rules, definitions, or implementations (tiering, targeting, buyer/bidder qualification, spam
+  heuristics, search ranking algorithms, data partnerships...)
+- Un-announced feature work
+- Industry or user research or analysis that may hold some competitive value
```
dleve123: Suggested rewording: “User research or Industry analysis that ...”. I think the double-or phrasing reads a bit weird.

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+- How to set-up a local development environment
+- Definitions of domain models
+- API documentation
+
+It _almost_ goes without saying, but other repo-specific interactions such as presenting the rationale for a design
+or offering code-review should leverage the built-in mechanisms of
+[commit messages](/playbooks/engineer-workflow.md#commits),
+[pull requests](/playbooks/engineer-workflow.md#pull-requests), and comments.
+
+## Public versus private
```
dleve123: Super minor: recommend titling “Public versus private content” - will make the document a bit more scannable IMO

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+When documentation relates to a particular system, leverage in-repository `README`s or `doc/`s wherever possible.
+This includes when the consumers may not exclusively be engineers.
+
+E.g.:
+
+- How to set-up a local development environment
+- Definitions of domain models
+- API documentation
+
+It _almost_ goes without saying, but other repo-specific interactions such as presenting the rationale for a design
```
dleve123: I’d recommend dropping the opening parenthetical statement, and starting this paragraph with “Other repo-specific ...”

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+- Security or compliance topics, including pending security work items or vulnerability-handling procedures
+- Incident/support instructions, including data recovery or incident-mitigation procedures
+- Sensitive business rules, definitions, or implementations (tiering, targeting, buyer/bidder qualification, spam
+  heuristics, search ranking algorithms, data partnerships...)
+- Un-announced feature work
+- Industry or user research or analysis that may hold some competitive value
+- References to private content (according to the criteria above) including documentation of private systems, links
+  to private gists that may include private data or logs, etc.
+
+## General engineering content
```
dleve123: I think we can do a better job tieing this section and the above together. I believe the goal of this section is to describe what belongs in README versus potential, thoughts on retitling to “artsy/README vs. artsy/potential”?

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+
+This public [artsy/README](https://github.com/artsy/readme) repo is our default location for shared engineering
```
dleve123: Thoughts on s/shared/public — you use the word public earlier in the doc for, I believe, as a synonym.

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+content such as:
+
+- onboarding resources
+- team procedures and playbooks
```
dleve123: I’d recommend adding “any other public content”

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+
+The private [artsy/potential](https://github.com/artsy/potential) repo supplements that with internal resources
```
dleve123: “That” is a bit unclear in this context, I’d recommend replacing with “artsy/README” instead.

Also, I’d s/internal/private so that we stay consistent with terminology.

playbooks/documentation.md
dleve123: I’d recommend adding a bullet for any other private content.

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Contact or role details for individuals
+- Information about candidates or new hires before they've joined or consented to announce their move
+- Production data or logs (including indirectly through migration output, etc.)
+- Repositories representing substantial business value, novelty, or criticality (such as the main API, content
+  management system, or auction operator tools)
+
+Content that _maybe_ should be private, depending on its sensitivity:
+
+- Security or compliance topics, including pending security work items or vulnerability-handling procedures
```
dleve123: I strongly disagree here. I think compliance and security information should always be private by default, and deliberately made public when appropriate.

This is aligned with the Priciple of Least Privilege - a well known and widely accepted security principle.

playbooks/documentation.md
```diff
@@ -0,0 +1,78 @@
...
+- Incident/support instructions, including data recovery or incident-mitigation procedures
+- Sensitive business rules, definitions, or implementations (tiering, targeting, buyer/bidder qualification, spam
```
dleve123: This is confusing to me. Sensitive business rules seems exactly like something that should be private as they almost always substantiate significant business value.

Can you clarify here?

playbooks/README.md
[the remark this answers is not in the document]
dleve123: Should this diff be reverted then? Can you elaborate a bit more on these create-readme scripts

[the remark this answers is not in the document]
orta: Nope, it'll just get overwritten again - docs for the repo's scripting are in [the contributing guide](https://github.com/artsy/README/blob/master/CONTRIBUTING.md#helpful-automation)

I'll need to add support for sub-folders, shouldn't block this PR.

playbooks/documentation.md
joeyAghion: Interesting! I think I got the opposite feedback about other forums, like the #security slack channel and trello board, which were invite-only. (In those cases we were debating between being public to the whole Artsy org and _not_ the whole world, but similar principles apply.)

I guess I meant to allow some space for more general security discussion without shunting them out of the repo to which they're relevant. E.g., if we were considering adding [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to asset references in Force, could that be discussed on a PR there?

playbooks/documentation.md
joeyAghion: Will remove "sensitive." Was just trying to distinguish the logic that _might_ be sensitive from the more mundane stuff.

playbooks/documentation.md
dleve123: I think there's a big difference between having security information being open within a private organization by default versus being open to the public by default, but point well taken. 

I do think it's generally safer to keep all security matters closed by default. In particular, I think it would be undesirable from a business perspective to have security incidents open as they progress. I would prefer to be less ambiguous, make a clear statement that, by default, these things should not be in the public-domain, than selectively promote such content to be public.

joeyAghion: Agreed. Moving this to the other section.

orta:
OK, I'm assuming this is now good as it looks like a lot of that conversation is classed as outdated.

PR #112, opened 2018-11-19
Remove data from open standup

orta:
This one is your call, you could opt to have someone else from inside the team give updates (which is what we do with product ATM) but it's cool either way 👍


cavvia:
Thanks for raising - I wasn't aware we were on the official agenda. We won't have an update every single week so let's take us off. I'll be in the meeting most but not all weeks and I can pipe up if we have anything to announce or share.

PR #114, opened 2018-11-21
Update Hokusai.md

izakp:
New release binary location - Hokusai v0.5.0 also fully supports credentials read from either environment or the ~/.aws/credentials file


izakp:
Merging so we can point devs to the new binaries in S3

PR #115, opened 2018-11-26
Mention the onboarding-eng Slack channel

### LINKED ISSUE #76: [RFC] New Hire Buddies
Insight: assigning someone as a mentor to a new hire that has been at Artsy for a while is good because they know lots of things, but it's bad because it's probably been a long time since they set certain things up. Another new hire is in a much better position to assist with these types of things because they just solved these problems!

For example, @pepopowitz just started and @starsirius asked if I'd like to be a mentor for him and I was happy to. I was able to help with a couple questions he had. Later he was asking about creating AWS and Jira accounts - something I haven't really thought much about.

@javamonn (for example) would be better suited to help getting someone setup on Jira since he's just done it and I actually don't even remember how I did it. Similarly, @ashleyjelks has just setup her AWS creds and ran into gotchas - she's in a much better position to help him out.

As a bonus, it feels good when you're new to a company to be asked to help with things. Makes one feel like they know stuff and are being helpful!

So, I wonder if there's a way to connect new hires so that they can learn from each other. It could be as informal as a slack group or formalized by assigning buddies in some way. Open to ideas here!

jonallured:
In an effort to connect new engineers, I've added a link the `onboarding-eng` Slack channel and nudge to have new hires :wave: at each other.

closes #76

### DIFF
--- a/.github/ISSUE_TEMPLATE/engineering-onboarding.md
+++ b/.github/ISSUE_TEMPLATE/engineering-onboarding.md
@@ -10,6 +10,7 @@ _Goal: I’ve met my manager and some members of the engineering team (over lunc
 google (artsymail), 1Password, and know how to use the coffee machine._
 
 - [ ] Add a cool pic of yourself to Slack and Jira
+- [ ] Say hi in the [onboarding-eng](https://artsy.slack.com/messages/CCC37HXE0) slack channel and give BLANK a :wave: as they are the most recently onboarded person.
 
 ## Your first week
 


alloy:
Probably should also add this to the list of default channels for people in the `@developers` Slack group.

jonallured:
@alloy good idea - where is this list?

dleve123:
I believe we moved onboarding to artsy/potential (my example onboarding issue: https://github.com/artsy/potential/issues/211), so this should be updated there.

jonallured:
@dleve123 oh hrm, very strange. Looks like the GitHub issue template is in README, but the actual onboarding issues are in potential? I wasn't aware of this change from onboarding in the open to moving it to potential. @orta any insight here?

orta:
The issue gets posted in potential, but the templates haven't moved and IMO probably shouldn't. I moved the issues as they have personal-information in them into Potential but the templates don't (and then they also live with the rest of the general on boarding).

orta:
The list is inside our slack group, I'll add it now

orta:
OK, cool 👍

Issue #117, opened 2018-11-27
[RFC] New dependencies to Emission/Reaction go through the RFC process


orta:
## Proposal:

Fail PRs which add new keys to the `dependencies` section on the `package.json` in Reaction / Emission. 

The fail would then be removed when an RFC is referenced in the body of the PR. Perhaps it canlook for a string like `RFC: [ur]` inside the body. Also support a message like `#skip_rfc` tag to skip the fail. 

## Reasoning

We'd like to find a better space to discuss additions to our two larger component libraries, it can be a shame to write a PR only to have it blocked on the discussion around adding a dependency

This is mainly to:

- Keep our dependency tree somewhat under control
- Ensure that we re-use dependencies, or avoid them if we have existing functionality you didn't know about
- Give everyone the chance to see and react to dependency updates

Some recent examples: 

 - [Formik in Emission](https://github.com/artsy/emission/pull/1063) (and yeah, I was a bit curt there, sorry @yuki24 (I think I was between meetings))
 - [React Spring](https://github.com/artsy/reaction/pull/1114)  

Slack:

 - [An SVG renderer for React Native](https://artsy.slack.com/archives/C02BAQ5K7/p1543269689301900) 🔒 
- [MapBox for React Native](https://artsy.slack.com/archives/D3PSUPD44/p1543350077001100)

## What an example RFC could look like

- [iOS Navigation Options](https://github.com/artsy/emission/issues/501)

## Exceptions:

Some PRs are pretty trivial, and so it should be possible to skip the discussion.

## Additional Context:

You can see our discussion [in notion](https://www.notion.so/artsy/27-11-2018-6161fc3144954d958c934a198e10087b) (it's a bit light on notes because I was the one taking notes and talking at the same time )

ashfurrow:
I like this. I'd rather swing too much in favour of process and then pull back, than to continue without enough.

damassi:
I'm into it too -- let us use Force's [`package.json`](https://github.com/artsy/force/blob/master/package.json#L52) as an example! 

@orta - Does this also apply to dev tooling / `devDependencies`? (I imagine this as only relating to deployable `dist` code.)

orta:
Not to `devDependencies`, no - only distributed code

javamonn:
We should do what we can to encourage that the necessary discussion happens upfront - failing a PR is probably too late as usually by that time your implementation is pretty coupled to the dependency that you introduced.

[...]

PR #123, opened 2018-11-29
Adds a doc on dependency management

orta:
Re: Our RFC on adding RFCs for new dependencies https://github.com/artsy/README/issues/117

This provides the general rubric for adding new dependencies to projects - we've wrote quite extensively on this in the past (see the notes section), so this doc aims specifically to focus on the down-side to dependencies and how to be cautious with them.

### DIFF
--- a/playbooks/dependencies.md
+++ b/playbooks/dependencies.md
@@ -1,0 +1,91 @@
+---
+title: Dependency Management Recommendations
+description: How do we want to handle new dependencies to our apps?
+---
+
+## Adding a Dependency
+
+In general we want to have a conservative dependency policy, which can be summed up as:
+
+- Is it worth the time it saves? (In terms of API surface area, runtime cost, complexity.)
+- Does it neatly fit the current problem domain, and are we likely to use the other parts?
+- If it becomes un-maintained, can we take over?
+
+Our policy should be **more** conservative when it’s code that is shipped to an end user or to multiple systems
+within Artsy. E.g. A dependency added to palette (our design system primitives) ends up in many repos transitively.
+
+This policy can be **less** conservative when it’s not used outside in production at runtime. Like, task runners,
+developer tools, linters or CI steps. Issues around these tools won’t block daily work and can be removed in
+exchange for some kind of tech debt.
+
+This document aims to emphasise the downsides of dependencies; but there are a lot of positives! We should aim for
+"Proudly found elsewhere" to ensure we don’t duplicate existing work and help improve projects for everyone. We
+want to be adding thoughtfully, not blocking any new dependencies.
+
+## Why?
+
+Dependencies help you go from `A -> B` faster, maybe `B` is a working in-app modal popup which acts like how you’d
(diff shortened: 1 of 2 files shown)


orta:
Note to self to add a link back in https://github.com/artsy/README/blob/master/culture/engineering-principles.md#own-your-dependencies when this is done

playbooks/dependencies.md
```diff
@@ -0,0 +1,91 @@
...
+- https://www.fngtps.com/2013/a-quick-note-on-minimal-dependencies-in-ruby-on-rails/
+- http://artsy.github.io/blog/2017/01/04/OSS-FAQ/
+
+Referential time-sinks, but valuable for extra info
+
+- https://guides.cocoapods.org/making/quality-indexes.html
+- https://gist.github.com/dominictarr/9fd9c1024c94592bc7268d36b8d83b3a
+- https://blog.npmjs.org/post/141577284765/kik-left-pad-and-npm
+
+[vendor]: https://en.wiktionary.org/wiki/vendor#Verb
```
javamonn: We should link to something that explains the usage of "vendor" specific to dependency management. I can't find a whole lot, but maybe something like [this](https://codeengineered.com/blog/2015/go-should-i-vendor/) would work.

playbooks/dependencies.md
```diff
@@ -0,0 +1,91 @@
...
+- Would this dependency be the first time we’ve needed something of this domain?
+- Do you feel well versed in the domain of this dependency and/or could maintain it if that needs to become an
+  option?
+
+## Ways to safely work around this
+
+- Re-use existing dependencies, or build the exact implementation that addresses just our needs today. Rather than
+  trying to foresee the future, we should re-asses once we have more information.
+
+- Directly import just the code you need, see [Importing directly](#importing-directly).
```
javamonn: I'd argue that vendoring is no different than introducing a dependency, except that it lives in your codebase - the same rubric above should still apply.

playbooks/dependencies.md
```diff
@@ -0,0 +1,91 @@
...
+- Are the dependencies of this dependency already in our projects, or is the dependency itself a transitive
+  dependency of another dependency we already rely on?
+- Would this dependency be the first time we’ve needed something of this domain?
+- Do you feel well versed in the domain of this dependency and/or could maintain it if that needs to become an
+  option?
+
+## Ways to safely work around this
+
+- Re-use existing dependencies, or build the exact implementation that addresses just our needs today. Rather than
+  trying to foresee the future, we should re-asses once we have more information.
```
javamonn: I'm curious if we can word this around ["Build for 10x"](https://github.com/artsy/README/blob/master/culture/engineering-principles.md#build-for-10x)

playbooks/dependencies.md
alloy: Great point! This section really just tries to suggest an option to get around some “no” answers to the rubrics, might need to be reworded to clarify that.

PR #126, opened 2018-12-02
RFC and retro now live in artsy/README

JuanitoFatas:
Use relative path, so in case of repo move again, links will still work 😉


alloy:
On fire 🔥

PR #145, opened 2019-01-11
Added #dev-ersity to the slack channel guide

zephraph:
We value having a diverse engineering team. Elevating where we have conversations around this is important, so let's show new folks where to find it.


orta:
yah, works for me, merge on green

PR #157, opened 2019-01-30
Adds to onboarding: general artsy/README callout, Engineer Workflow doc

mdole:
I found the Engineer Workflow doc very helpful for better understanding how we use GitHub and other services. Also noticed that there were a lot of links to artsy/README docs, but no general link to artsy/README itself - thought it worth calling out.


orta:
I don't think the rendered markdown works for this change

mdole:
Ooo you are correct, sorry about that. will fix and update

mdole:
Merp 4 spaces instead of 2 T_T fixed, thanks @orta

orta:
👍

PR #158, opened 2019-02-05
Adding more resources to retrospective playbooks

SamRozen:
Adding here links shared in other channels.
- 'Spotify Retro Kits' has a lot of interesting formats.
- 'How we made out team retros more effective' is an interested read about how important it is to be result oriented.


sweir27:
Awesome, thank you!

PR #160, opened 2019-02-06
Documented migration from Heroku to K8.

dblock:
The PR in Doppler is quite detailed about all the things I had to do with @ashkan18's help.


damassi:
👍 👍

PR #164, opened 2019-02-18
Update docs for trading shifts

pepopowitz:
There is already a note in the "Scheduling" playbook, but I added one to the general "Support" playbook.


sweir27:
Thank you!!

PR #173, opened 2019-03-26
@damassi: Document that responsive page designs are [I think?] preferred

joeyAghion:
I think we "decided" broadly that we want to implement pages in a responsive manner to provide a first-class experience to the growing and changing mobile audience. I thought I'd document that for posterity.

(Please ignore the prettier changes... I guess there were offending commits pushed earlier.)


damassi:
👍 Thank you!

PR #180, opened 2019-04-05
Replace artsy/meta with artsy/README

dleve123:
artsy/meta is deprecated and artsy/README is its spiritual successor


orta:
Ah yeah, great point!

PR #193, opened 2019-04-30
Suggest that people add Github and AWS usernames to their Slack profiles

joeyAghion:
Minor addition to onboarding checklist: suggest that people list their Github and AWS usernames on their Slack profiles. (Everything else is just `prettier`.)

This helps with [PLATFORM-1225](https://artsyproduct.atlassian.net/browse/PLATFORM-1225) and I've already added these fields to our org's profile template.


mdole:
:+1:

PR #195, opened 2019-05-01
Update engineering workflow document with new PR role definitions

mdole:
Following up on [RFC 177](https://github.com/artsy/README/issues/177), this PR updates documentation around PR standards at Artsy. It will be followed by a pair of Peril rules, one that warns if > 1 person is assigned to a PR, and one that adds a comment 1 business day after someone is requested to review if they have not yet reviewed.

Let me know if you have any feedback!

PR #197, opened 2019-05-02
Improve language around what code review + assignment entail

mdole:
@dblandin noted that the initial updates made [to this doc](https://github.com/artsy/README/pull/195)  included an unhelpful definition of code review - for now I've simply generalized the code review requirements, and I think a future RFC around code review is an excellent idea.

Thanks for the thoughtful feedback!

cc @joeyAghion - thanks for helping to clarify the assignee/reviewer designation!

### DIFF
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -82,7 +82,7 @@ Once again, the title should concisely explain the change or addition. The descr
 **_Request a review_** from at least one team member. This should be someone who has context around the changes
 that were made, experience with the system(s) being modified, and/or knowledge of the language/framework in use.
 Team members that are not explicitly assigned to a PR are encouraged to contribute as well, but reviewers have the
-explicit responsibility to help make sure the code is elegant, understandable, and will not have unforseen impacts.
+explicit responsibility to read the code in question and provide commentary or feedback on it.
 
 If you want to let someone know the PR is happening without requesting their review, just @ them in the body of the
 PR.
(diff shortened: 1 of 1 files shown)


playbooks/engineer-workflow.md
```diff
@@ -91,9 +91,13 @@ Reviewers are asked to leave their review within one business day of being reque
...
-**_Assign_** another team member (and only _one_ team member) to the PR. This is the person who is responsible for
-moving the PR forward. Once someone has been assigned to a PR, they should make a good faith effort to take one of
-the following actions within one business day:
+**_Assign_** a single team member (they will often also be one of your reviewers) to the PR. This is the person who
+is responsible for moving the PR forward.
+
+Assigning exactly one reviewer - no more and no less - helps ensure that there's no question about who is
+responsible. Since assignees are not explicitly expected to review, it's perfectly fine to both assign and request
+a review from the same person. Once someone has been assigned to a PR, they should make a good faith effort to take
+one of the following actions within one business day:
```
damassi: If the `Assignee` is responsible for moving the PR forward, then it should be expected that they're responsible for line-by-line reviewing. 

With the `Assignee` being responsible for moving a PR forward, any number of `Reviewers` can then be assigned as additional sets of eyes. Often times a PR covers a lot of domain territory that involves different expertise. Why would be want to explicitly define how feedback is received in this way?

@joeyAghion noted earlier that often times _too_ many reviewers can lead to conflicting requests for changes. However, in practice, it's rare that this happens as typical PRs are dev -> assignee, without additional reviewers. When additional reviewers are requested, this is typically an action taken by the developer who is specifically looking for additional feedback. This shouldn't be discouraged.

mdole: @damassi I'm not sure I follow - I don't see this as limiting the amount of reviewer feedback or discouraging additional review requests. Help me understand where you're coming from?

damassi: This bit of language: 

> Assigning exactly one reviewer - no more and no less - helps ensure that there's no question about who is
responsible

As well as this: 

> Since assignees are not explicitly expected to review

damassi: Ah, reading through the context I see this is clarified by sub-headings, all good 👍

PR #199, opened 2019-05-06
Removes CTO reference

ashfurrow:
Given dB's departure, this PR updates the standup template.


oxaudo:
Or use your common sense while reading from the screen:) Thanks!

PR #201, opened 2019-05-06
[Stand-Up] Add prompt for each team's update

SamRozen:
Explicitly add prompt to each TL to source team's milestones during standup.

### DIFF
--- a/events/open-standup.md
+++ b/events/open-standup.md
@@ -78,7 +78,7 @@ _Cross-dependencies / Requests for Pairing_
 
 -
 
-_New Milestones / Repos / Blog posts_
+_New Milestones / Repos / Blog Posts / New features or updated functionality released: prompt Auction, Gallery, Platform, Grow, Discover, Purchase teams_
 
 -
 


events/open-standup.md
```diff
@@ -78,7 +78,7 @@ _Cross-dependencies / Requests for Pairing_
 
 -
 
-_New Milestones / Repos / Blog posts_
+_New Milestones / Repos / Blog posts (prompt each TL for team's milestone: Auction, Gallery, Platform, Grow, Discover, Purchase)_
```
oxaudo: I would maybe phrase this as New Milestones / Repos / Blog Posts / New features or updated functionality released: prompt Auction, Gallery, Platform, Grow, Discover, Purchase teams for those In this proposed version seems like this prompt in () is part of Blog Posts.

jonallured:
LOL, I love when things come full circle! 😝 

@SamRozen with the history here, I'd suggest turning this PR into an RFC. To be honest, I don't miss TL updates at all and in fact feel like standup is starting to go a little long as it is. I still maintain that I get this info from multiple vectors and hearing it again at standup is needless repetition. Would love to hear from others though! ❤️

ashfurrow:
@jonallured Yeah, I wouldn't want to go _back_ to the team updates we had. This PR seems to be something different: what has the team shipped recently? I would frame it not as a team update, but a celebration of recently-accomplished work.

SamRozen:
@jonallured @ashfurrow my intent with keeping this at the end is that it can be a quick update on major achievements. I'd like us as an engineering org to spend more time celebrating the business impact we have. And I definitely agree that a full length status update from each team is not useful. Happy to move to RFC format if you think it's a controversial change.

jonallured:
Sorry, one more thought occurred to me! 😝 

What I like about the updated language in this PR is that it moves the prompt from Tech Lead to Product Teams, which I think is nice. There's no reason leads need to be the source of these updates. I'd probably update the PR title to reflect this change too:

```diff
+ [Stand-Up] Add TL prompt for each team's update
- [Stand-Up] Add prompt for each team's update
```

ashfurrow:
@jonallured agreed! This changes encourages ICs who worked on a feature announce it themselves. I'm going to update the PR title and merge. Thanks again everyone!

PR #198, opened 2019-05-03
Start a custom "technology radar" for Artsy

joeyAghion:
For a while we've talked about making shared technology decisions and strategy more visible across Artsy. This is a quick attempt to build a custom "technology radar" for Artsy using [ThoughtWorks' approach and tool](https://www.thoughtworks.com/radar/faq).

There are lots of gaps and I'd love for others to add the significant technologies they're discouraging, evaluating, and encouraging to this framework.

The links should start working after this is merged 🤞, but in the meantime you can interact with a version built with the same data [here](https://radar.thoughtworks.com/?sheetId=https%3A%2F%2Fgist.githubusercontent.com%2FjoeyAghion%2Fde544d14e32778facd5e96fd154aaff5%2Fraw%2Fartsy_technology_radar.csv). Screenshots:

Some orgs give formal ownership of this process to a technology review committee and reevaluate status every quarter or month. It's interesting to imagine how we might evolve a process like that.

PR #202, opened 2019-05-07
[Tech Radar] Add Relay, Redux, and Fastlane

alloy:
https://github.com/artsy/README/pull/198#issuecomment-489619450


playbooks/technology_radar/data.csv
```diff
@@ -46,3 +46,6 @@ APIs as a product,assess,techniques,TRUE,
 GraphQL batching,assess,techniques,TRUE,
 Bug bounties,hold,techniques,FALSE,
 Serverless architecture,hold,techniques,TRUE,
+Relay,adopt,languages & frameworks,FALSE,
+Redux,assess,languages & frameworks,FALSE,
```
alloy: @damassi While we’re using this in production (eg Prediction), with the newer React hooks API I _think_ the understanding is that in the future we still need to assess if we ever need to add full Redux again or if the builtin API is enough (like has been done in the new nav bar iirc). Is this correct?

alloy:
Ah, I was unsure where to add them and had to resist the urge to sort them alphabetically, so ended up appending.

alloy:
@joeyAghion Ok, I added a sort script because I don’t want to think about this any further nor have others need to do so 😎

joeyAghion:
Sorting alphabetically _definitely_ won't work. The visualizer will order the concentric rings based on the order in which those labels appear, as I found out here:



(I'm sorry for the trouble but hopefully my headaches can save you some. I tried to describe this in the README but it could use some clarification.)

Personally, I'd remove the script and implicitly encourage the right ordering by keeping the file consistent. If you like having the script, it just needs to ensure that "adopt," "trial," "assess," and "hold" appear in that order.

playbooks/technology_radar/data.csv
damassi: @alloy - I'm of the opinion that for anything app-like that contains significant amounts of state we should continue using Redux (Prediction, Position). It's patterns scale very well, the ecosystem is vast, and so much comes out of the box for free (such as performance tuning, middleware pipelines, devtools, etc). There's some really nice hooks-centric, typescript-friendly Redux libs out there, such as [easy-peasy](https://github.com/ctrlplusb/easy-peasy). 

For Most Artsy webpage-like code (think, Force "app shell", new nav bar, etc) hooks + context API is more than enough.

alloy: That does seem to kinda follow my outline. I chose ‘asses’ to mean “do not use it by default, but if we determine that an app definitely benefits from a fuller redux-like env, only then do we use it”. But perhaps that should be ‘adopt’ and it’s implicitly clear that even for tech in ‘adopt’ we need to evaluate well on a case by case basis? (cc @joeyAghion)

alloy:
Well that’s a little disappointing. I’m not sure I understand why their tooling doesn’t just sort things the intuitive way… 🤷‍♂ However, I don’t think we need to think about it any further, so I updated the script.

playbooks/technology_radar/data.csv
damassi: I think 
> Redux,assess,languages & frameworks,FALSE,

Is accurate 👍

alloy: Done 👍

PR #209, opened 2019-05-16
Update on-call schedule docs to reflect OpsGenie system

dblandin:
We've recently switched to OpsGenie for scheduling on-call rotations. This PR updates our documentation to include scheduling steps and instructions to override shifts.


playbooks/support/scheduling.md
```diff
@@ -1,59 +1,74 @@
...
-   like:
-
-```
-Hi team! Just finished creating calendar invites for this round of on-call. If something looks wrong, please let me know! Soon I will be deleting the Notion doc so we can use this as the source of truth. Note that you can still switch shifts if something comes up!
-```
-
-9. After a couple of days, delete the Notion doc to avoid confusion.
-10. Schedule a reminder to repeat this process for the next round.
+   to begin (can use [Team Navigator][] as a reference)
+1. Randomly split engineers into two groups (A and B)
```
sweir27: Clarify that all non-HQ engineers should be in the same group

playbooks/support/scheduling.md
```diff
@@ -1,59 +1,74 @@
...
+   <img width=500 src="images/opsgenie-schedule-5.png" />
+   <img width=500 src="images/opsgenie-schedule-4.png" />
+   <img width=500 src="images/opsgenie-schedule-3.png" />
+   <img width=500 src="images/opsgenie-schedule-0.png" />
+
+1. Create a new OpsGenie rotation (2 of 2)
+
+   - Add all participants from group B into the rotation
+   - Set the rotation type to "weekly"
+   - Set the "Start on" date to correspond with the "End on" date from the previous second rotation
```
sweir27: Clarify that one of these will start on a Monday and one on a Wednesday

PR #213, opened 2019-06-05
update Kubernetes docs about draining nodes when they fail to schedule pods

izakp:
Add some docs about configuring Kubernetes with admin credentials via IAM auth, and an example of identifying, draining and terminating a node with an unresponsive Docker daemon.

@eessex I based this off a [screenshot](https://artsy.slack.com/archives/CA8SANW3W/p1559675210026500?thread_ts=1559674010.024300&cid=CA8SANW3W) you took of a Pod in a bad state in the dashboard, but can't remember how to link images in this repo


eessex:
Awesome, I think this makes sense without screenshots -- thank you!

PR #214, opened 2019-06-05
Add instructions on how to update environment variables

sweir27:
#minor but I didn't see instructions for this anywhere!


izakp:
Thanks @sweir27 !

izakp:
FWIW this is mentioned in https://github.com/artsy/hokusai/blob/master/docs/Getting_Started.md but not a bad idea to duplicate

PR #220, opened 2019-06-27
Add running peer learning groups to open standup

xtina-starr:
This adds a section to mention the running peer learning groups during engineering open standup.


ashkan18:
YESS!

PR #221, opened 2019-07-02
add Cloudflare playbook

izakp:
Adds a Cloudflare playbook

### DIFF
--- a/playbooks/cloudflare.md
+++ b/playbooks/cloudflare.md
@@ -1,0 +1,32 @@
+---
+title: Cloudflare
+description: Working with Cloudflare
+---
+
+# Cloudflare
+
+## Cloudflare setup
+
+Cloudflare is configured via Terraform in the [Infrastructure repo](https://github.com/artsy/infrastructure/blob/master/terraform/staging/cloudflare.tf)
+
+Manage new protected domains by making PRs to this configuration.
+
+## Cloudflare DNS
+
+To route requests through Cloudflare's firewall...
+
+1) A [DNS record for artsy.net](https://dash.cloudflare.com/0373426be7be649ff052277fb5377c4f/artsy.net/dns) is created in Cloudflare's interface providing the CNAME of the ELB backing the given service.  You can obtain the DNS record of the ELB for a service by running `hokusai [staging|production] status` and looking at the `EXTERNAL-IP` column for the service record.
+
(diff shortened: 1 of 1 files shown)


eessex:
We have some docs on cloudflare here, https://github.com/artsy/potential/wiki/Platform-FAQ#enabling-ddos-protection-for-a-subdomain-via-cloudflare
It might make sense to combine these or cross-link them.

playbooks/cloudflare.md
```diff
@@ -0,0 +1,34 @@
...
+
+3) The value `loadBalancerSourceRanges` is set on the [service ELB](https://github.com/artsy/force/blob/master/hokusai/production.yml#L143) restricting traffic coming from Cloudflare's edge nodes
+
+With this setup, Cloudflare resolves requests to IPs of their edge nodes, then makes requests to our ELBs as origin servers.
+
+## Routing requests around Cloudflare
+
+In the case of Cloudflare edge node failures, we have the option to route requests to pass through the Cloudflare firewall on its edge nodes.  In the DNS settings, if the domain's "Status" field with the *☁* icon on the right-hand column is orange and reports "DNS and HTTP Proxy (CDN)" then the firewall is active and DNS requests for `*.artsy.net.cdn.cloudflare.net.` will resolve to Cloudflare edge node IPs, as is described above.
+
+However, this can be toggled to a greyed-out *☁* icon that reports "DNS only" and will serve DNS requests for the domain `*.artsy.net.cdn.cloudflare.net.` as a CNAME for the origin server.  In this case, requests resolve to the ELB's IP addresses and do not pass through the Cloudflare edge nodes.
```
izakp: Technically there's a 2x CNAME resolution happening here, tho not sure how deep we want to go in the case (like today) we can't access Cloudflare's dashboard... so in the case off `www.artsy.net` resolves to a `CNAME www.artsy.net.cdn.cloudflare.net.` served by Dyn, which then resolves to either the Cloudflare edge node IPs (DNS / Http proxy enabled) or `CNAME a9a61bf76c24511e8a1960ea9b7c2ad2-1501559720.us-east-1.elb.amazonaws.com` (DNS only)

joeyAghion:
I agree that this should be combined with the items in the Potential wiki, since that includes our support-oriented playbooks and this seems to match.

PR #225, opened 2019-07-10
@zephraph: Clarify that release PRs appreciate titles

joeyAghion:
Lately many of our release PRs are just called the default `Staging`, so let's encourage more deliberate  choices.

(The other stuff is prettier.)


zephraph:
Cool, looks good!

PR #228, opened 2019-07-12
Update on-call docs based on our transition to OpsGenie

pepopowitz:
I've had to raise a couple incidents this week, and noticed that we didn't have all our README docs updated from our transition to OpsGenie. This PR addresses that.

### DIFF
--- a/playbooks/support/responding.md
+++ b/playbooks/support/responding.md
@@ -39,8 +39,6 @@ respond to these issues.
 
    Any incident with public impact should be communicated externally via the status page.
 
-   - Use your judgement to select an "Impact" based on severity and duration of the incident. Refer to the
-     [section on Severity of Incidents for guidance](#severity-of-incidents).
    - Communication does not have to be detailed. Not every stage of the incident needs to be communicated on status
      page.
    - Remember that this page is targeted to a public audience (i.e., users and partners) of Artsy products. Some
@@ -128,8 +126,6 @@ Critical incidents have a very high impact:
 - Customer data loss.
 - “All hands on deck.”
 
-The **Critical** level in OpsGenie equates to the **Critical** impact level in statuspage.io.
-
 #### Examples
 
 - Artsy website is down.
@@ -142,8 +138,6 @@ High incidents have a significant impact:
 - A user-facing service is down for a subset of users.
 - Core functionality is impacted for a specific use-case.
 - Need to find the best people to solve as soon as possible.
-
-The **High** level in OpsGenie equates to the **Major** impact level in statuspage.io.
 
 #### Examples
 
@@ -159,8 +153,6 @@ Moderate incidents have a low impact but require some urgency:
 - Core functionality is impacted for a single resource.
 - Should inform the relevant product team about the issue.
 
-The **Moderate** level in OpsGenie equates to the **Minor** impact level in statuspage.io.
-
 #### Examples
 
 - Broken links in Eigen on UNTITLED art fair page
@@ -173,8 +165,6 @@ Low incidents have low impact and low urgency:
 
 - Known issues that are on a product team’s radar.
 - Issues that can be handled in a regular sprint’s work.
-
-Low incidents do not require an update to the Artsy status page.
 
 #### Examples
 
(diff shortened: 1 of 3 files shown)


playbooks/support/responding.md
```diff
@@ -75,28 +94,93 @@ One of the on-call members handles internal and external communication for the i
...
+### P1 - **Critical**
+
+Critical incidents have a very high impact:
+
+- A user-facing service is down for all users.
+- Confidentiality or privacy is breached.
+- Customer data loss.
+- “All hands on deck.”
+
+The **Critical** level in OpsGenie equates to the **Critical** impact level in statuspage.io.
```
pepopowitz: This one lines up pretty well, but the rest of the levels I guessed at. I decided on...

OpsGenie : StatusPage
Critical : Critical
High : Major
Moderate : Minor
Low : (none, doesn't warrant a public-facing incident)
Informational : (none, doesn't warrant a public-facing incident)

Let me know if you think I got these wrong!

playbooks/support/responding.md
```diff
@@ -15,48 +15,67 @@ respond to these issues.
 ## Step 2: Raise
 
 1. When an incident is raised in the channel or identified in another way, the on-call team first **tracks the
-   incident** by creating a ticket in [Jira Ops](https://artsyproduct.atlassian.net/projects/INCIDENT/incidents),
-   based on the information they know at that time. Some workflow steps:
-
-   - In the description on the ticket, link to the slack thread in #incidents or #alerts.
+   incident** by creating a ticket in [OpsGenie](https://artsy.app.opsgenie.com/incident/list), based on the
```
dblandin: We should rename "ticket" to "incident" here.

playbooks/support/responding.md
```diff
@@ -15,48 +15,67 @@ respond to these issues.
...
+   - Enter a quick description of the problem as the Incident message.
+   - Enter a link to the #incidents thread as the Incident description.
+   - Select a Priority level from the list using your gut instinct. Refer to the
+     [section on Severity of Incidents for guidance](#severity-of-incidents).
    - Take your best guess at the other information, it can always be edited later.
 
-2. Using the slack integration on the Jira Ops ticket, create a new slack channel to discuss resolution.
+2. Create a Slack channel for this incident for discussion of the technical resolution.
+
+   - Name it #incident-N and invite the other on-call engineer and any other active responders.
```
dblandin: Should we describe what `N` represents or do you think it's intuitive enough?

playbooks/support/responding.md
dblandin: We might want to leave the mappings off for now. During the creation of an incident on StatusPage.io, you can't actually set the severity (which is only exposed to viewers of the page via the incident headline text color).

Instead, you set the status of components in your system and a severity is calculated for the incident based on the component statuses.

After creating the incident, you may override the calculated severity.

I think we owe some documentation on setting the appropriate status for components but we might want to stick with the calculated severity on StatusPage.io for now. What do you think?


source: https://help.statuspage.io/help/component-status-incident-impact-and-top-level-status-calculations

pepopowitz: Sure, that works. I'll remove those lines (and any references to them).

playbooks/support/responding.md
pepopowitz: I can add some clarification.

playbooks/support/responding.md
```diff
@@ -75,28 +93,85 @@ One of the on-call members handles internal and external communication for the i
...
+
+#### Examples
+
+- Charge details for a partner need updating.
+- Artwork appearance on the site (“how come it’s not showing up in…”).
+
+### P5 - **Informational**
+
+Informational incidents are used primarily for training. Use this for incidents you create to practice incident
+response.
```
dblandin: I wonder if we can be a bit less specific with this severity level.

I found this definition from Norton:

> These are routine security incidents with no impact to the client. They are presented for informational/reporting purposes.

source: https://us.norton.com/online-threats/glossary/s/security-incident-severity--informational.html

Maybe we can use something like:

Informational incidents have no customer impact. They are often created for reporting or training purposes.

pepopowitz: Good call, I like it. Changed!

dblandin:
Awesome! Going to go ahead and merge this in. We can continue iterating. :+1:

PR #230, opened 2019-07-15
Add docs for Volt-specific review app

jonallured:
This doc is what I've been using to create review apps for Volt and I'm trying to find a better home for it. Over on https://github.com/artsy/hokusai/pull/172 I got the feedback that this is a better place for it so I'm opening this to get it added.

### DIFF
--- a/playbooks/volt-review-app.md
+++ b/playbooks/volt-review-app.md
@@ -1,0 +1,75 @@
+# Volt Review Apps
+
+The authoritative source of information for creating review apps using Hokusai
+can be found here:
+
+https://github.com/artsy/hokusai/blob/master/docs/Review_Apps.md
+
+But I've found that a bit noisy and so I'm distilling all that info to a short
+list of commands to run:
+
+```sh
+# create a new config file
+$ hokusai review_app setup 1234
+# update config file with the image tag 1234
+$ sed -i "" "s/staging/1234/g" hokusai/1234.yml
+# commit the config file
+$ git add hokusai/1234.yml
+$ git commit -m "Config for review app 1234"
+# push the image - this takes a while...stop if it doesn't exit cleanly!
+$ hokusai registry push --force --tag 1234
+# create a review app called 1234
+$ hokusai review_app create 1234
+# copy over the env from staging
+$ hokusai review_app env copy 1234 --configmap nginx-config
+# inspect review app and get url
+$ hokusai review_app status 1234
+# or if you're feeling super 1337:
+$ hokusai review_app status 1234 | grep service | tr -s [:blank:] | cut -d ' ' -f 4 | pbcopy
+```
+
+## Update Gravity
+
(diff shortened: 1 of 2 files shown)


playbooks/volt-review-app.md
```diff
@@ -0,0 +1,75 @@
...
+$ hokusai registry push --force --tag 1234
+# create a review app called 1234
+$ hokusai review_app create 1234
+# copy over the env from staging
+$ hokusai review_app env copy 1234 --configmap nginx-config
+# inspect review app and get url
+$ hokusai review_app status 1234
+# or if you're feeling super 1337:
+$ hokusai review_app status 1234 | grep service | tr -s [:blank:] | cut -d ' ' -f 4 | pbcopy
+```
```
starsirius: I was imagining we can contribute the short version back to the canonical hokusai review app doc to help streamline dev's workflow. Is there anything special about Volt's use case?

I'm 👍 to start from this specific example and iterate, so definitely not blocking!

playbooks/volt-review-app.md
```diff
@@ -0,0 +1,75 @@
...
+$ sed -i "" "s/staging/1234/g" hokusai/1234.yml
+# commit the config file
+$ git add hokusai/1234.yml
+$ git commit -m "Config for review app 1234"
+# push the image - this takes a while...stop if it doesn't exit cleanly!
+$ hokusai registry push --force --tag 1234
+# create a review app called 1234
+$ hokusai review_app create 1234
+# copy over the env from staging
+$ hokusai review_app env copy 1234 --configmap nginx-config
```
izakp: Going to need another step here to copy Nginx config `hokusai review_app env copy 1234 --configmap nginx-config`

playbooks/volt-review-app.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Pushing updated images
+
+Time passes, you get feedback on your PR and you want to update your review app,
+no problem, just push a new image to the registry:
+
+```sh
+# push the image - this takes a while...stop if it doesn't exit cleanly!
+$ hokusai registry push --overwrite --skip-latest --force --tag 1234
+# redeploy the review app with the new image
```
izakp: As in https://github.com/artsy/hokusai/pull/172 I think it's worth noting that this will run pre-deploy and post-deploy hooks, so if your review app has a migration step and connects to the app's same staging database, it may have unintended consequences!

playbooks/volt-review-app.md
ashkan18: @izakp we may need a command for copying `nginx-secret` as well.

izakp: Ah yes -good point @ashkan18

jonallured:
Yeah, I've lost steam on this one - I too think that this belongs in Volt so I'm going to close this and we can revisit later if needed.

PR #234, opened 2019-07-31
Update mapping between Engineering Ladder and Artsy Levels

SamRozen:
We remapped the Engineering Ladder to Artsy's internal level.
This now reflects the accurate mapping.


sweir27:
Thanks for the speedy update!

PR #235, opened 2019-07-31
Converts capabilities into a matrix form

sweir27:
This will allow us to deprecate the duplicative notion doc!

I used html because the markdown syntax was getting pretty unwieldy given how involved the lines are.

### DIFF
--- a/careers/ladder.md
+++ b/careers/ladder.md
@@ -19,21 +19,20 @@ engineers.
 
 ## Tracks
 
-The Engineering Ladder encapsulates two tracks—an individual contributor (IC2–IC8) and manager track
-(M3–M6). The manager track is available to engineers who have achieved an "IC5 Senior Engineer" level on the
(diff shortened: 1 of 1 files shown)


careers/ladder.md
```diff
@@ -64,178 +63,212 @@ Scope of impact:   Self   >   Team   > Teams > Org
...
-
-- Primarily acts as a multiplier by building systems, authoring tools, or introducing policies or patterns.
-
-#### Discretion
-
-- Ensures Artsy makes judicious technology choices in the face of rapid business and technology advancements.
-- Identifies technology strategic growth opportunities that enable Artsy to grow as a business.
+### Capabilities
+
+<style>
```
SamRozen: @sweir27 actually Github doesn't seem to process your style tag properly.

sweir27: Good call, fixed in a separate PR.

PR #244, opened 2019-08-27
Remove orta :cry:

just-be-dev:
Noticed a left over reference to Orta when trying to schedule a lunch and learn.


oxaudo:
I think we could have still let him live for several years. I still find references to Craig sometimes -:)

Issue #245, opened 2019-09-05
[RFC] Create a technical plan before starting work on projects


zephraph:
## Proposal:

Before beginning work on a project, we should have a dedicated technical plan. This isn't to say that we aren't already doing planning, but rather an effort to make our planning more consistent, deliberate, and discoverable. 

At a high level the plan could include:

- Motivation (why are we doing the work?)
- Impact (who or what is impacted by the work?)
- Approaches (how will we do the work?)
- Risks (what should we watch out for?)
- Decisions (what do we need to decide on to progress?)
- Outcome (what does success look like?)

Technical project plans _should_ be stored in a similar location that make it easy to find from team to team. My recommendation is in notion under [Product] > [<team>] > [Technical Plan]. 

## Reasoning

- Assumptions may form that a certain part of a project was thought out or discussed but it may not actually have been. Writing it down makes it explicit.
- Placing project plans in a consistent, searchable space makes it easier for people to discover historical decisions and keep in tune with on-going work
- It'll force us to cover our bases

## Exceptions:

If a single ticket is sufficient, you probably don't need a plan. If the work takes more than a sprint, you definitely _do_ need one. Keeping with our principle of [minimal viable process](https://github.com/artsy/README/blob/master/culture/engineering-principles.md#minimal-viable-process) we only want to do a thing when it adds value. Use your best judgement. 

## Additional Context:

Here's the [slides](https://docs.google.com/presentation/d/1EpvkcmabLy-bok9WRsq3ArBYJcsjYau0f_Ul5G_h06Q/edit?usp=sharing) from our mini-conf on this topic

## How is this RFC resolved?

We agree to start creating technical plans and storing them in a particular place. 

---

If you have ideas of how to improve this proposal, please let me know!

ashkan18:
awesome!  Im thinking how can we make the project part of 👇 broader
> Before beginning work on a project
I think anytime someone wants to make decision between few possible solutions/options or they are building things that impact other teams and maybe other cases they should consider technical plan. 
I've personally used this in the past to make decisions with others helps and get their opinion on things that could technically be considered smaller than a project.

sweir27:
This is great! I'm excited to formalize this/get started.

One minor note-- an initial step after this RFC is resolved could be to co-locate the existing technical plans as a way to seed the effort, + create a template that can be easily extended.

dleve123:
A couple of thoughts. I'd encourage us to: 

[...]

PR #246, opened 2019-09-11
[RFC] Document expectations for systems of varying criticality

joeyAghion:
The platform practice discussed and documented some expectations we'd like to adopt going forward for systems of varying "criticality" ([notes here](https://www.notion.so/artsy/Criticality-framework-expectations-069b662c599e4aff849c0d0cdb661c05)). These should continue to evolve, but based on the general agreement in that meeting, I think it's worth getting feedback from the wider engineering team. Specifically:
* Are these reasonable expectations to set for our projects?
* Are there other standards we should try to enforce as systems gain business-critical usage?

This PR replaces our placeholders with real proposed standards. Some still-pending items are marked with `TODO`.

As usual the [full project list](https://www.notion.so/artsy/17c4b550458a4cb8bcbf1b68060d63e6?v=2b874803f96c483bbdc2e80b7fbd25f9) is a good reference for what systems exist and how critical they are.

(I highly encourage reviewing [the formatted version](https://github.com/joeyAghion/README/blob/playbooks/playbooks/system-criticality.md) of this doc.)


playbooks/system-criticality.md
```diff
@@ -31,21 +31,24 @@ Retired systems, spikes, or time-bound experiments that aren't significantly use
...
+| ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| Development       | READMEs contain up-to-date set-up instructions<br/>Code review on all [pull requests](/playbooks/engineer-workflow.md#pull-requests)<br/>New systems or significant new components undergo [tech review]() | &larr;ditto                                              | &larr;ditto +<br/>Production environment is replicable locally (TODO)                                                                   |
```
joeyAghion: @eessex you mentioned this concern about having high-fidelity local (or staging) environments available. Is there better language we could use here, or an example?

joeyAghion: Also, it would be nice to have a link that elaborates on what "technical review" means here. A technical plan format is proposed in https://github.com/artsy/README/issues/245, and that overlaps with a technical design review (e.g., it can serve as a starting point for a review discussion).

playbooks/system-criticality.md
```diff
@@ -31,21 +31,24 @@ Retired systems, spikes, or time-bound experiments that aren't significantly use
...
+| Monitoring        |                                                                                                                                                                                                            | Application instrumentation (e.g., Datadog)              | &larr;ditto                                                                                                                             |
+| Incident handling | Standard [incident response times](/playbooks/support/on-call-slo.md) (pending improved notifications)                                                                                                     | &larr;ditto                                              | &larr;ditto +<br/>Downtime automatically reported to Opsgenie as incidents<br/>Incidents are reported and updated on public status page |
```
joeyAghion: We discussed adopting the same incident response-times for all levels of systems, but does that truly match our reality? In practice many [level 1](https://www.notion.so/artsy/17c4b550458a4cb8bcbf1b68060d63e6?v=cb0164f2f5e94962872d8cf7a4039732) systems aren't part of critical workflows and minor incidents may get punted to backlogs.

playbooks/system-criticality.md
```diff
@@ -31,21 +31,24 @@ Retired systems, spikes, or time-bound experiments that aren't significantly use
...
+| Development       | READMEs contain up-to-date set-up instructions<br/>Code review on all [pull requests](/playbooks/engineer-workflow.md#pull-requests)<br/>New systems or significant new components undergo [tech review]() | &larr;ditto                                              | &larr;ditto +<br/>Production environment is replicable locally (TODO)                                                                   |
+| Testing           | Automated tests<br/>Tests are run on PRs and `master` by a Continuous Integration pipeline                                                                                                                 | &larr;ditto                                              | &larr;ditto +<br/>Test coverage tooling                                                                                                 |
```
SamRozen: Re `Test coverage tooling` do we expect these services to have a target test coverage? Are there generic rules that apply to all our level 3 services such as: must have at least 50% test coverage and no commit can decrease test coverage?

playbooks/system-criticality.md
```diff
@@ -31,21 +31,24 @@ Retired systems, spikes, or time-bound experiments that aren't significantly use
...
+| Availability      | External availability monitoring                                                                                                                                                                           | &larr;ditto                                              | &larr;ditto                                                                                                                             |
+| Performance       |                                                                                                                                                                                                            |                                                          | Latency-based monitors and alerting (tailored to service)                                                                               |
```
SamRozen: For level 3 services, is there a latency threshold that applies to every service? Should we have every service with p99 < 1s for instance ?

playbooks/system-criticality.md
joeyAghion: We discussed specifying a coverage ratio but decided not to since [at least for the time being] a single standard wouldn't be meaningful given the very different places projects are starting from.
Personally, yeah I'm hope that making coverage more _visible_ will give us some more experience and insight into the most practical additional standards to set, like requiring that commits don't decrease coverage.

eessex: We currently have a check on Force and Positron to target 70% coverage on a diff. There is a circle check that will fail if the diff is lower, but it does not block merging -- still working to tune this to specific app directories and ignore config files, which I think may confuse our stats. 
Screen Shot 2019-09-11 at 2 24 28 PM

playbooks/system-criticality.md
joeyAghion: I'd originally suggested <200ms average response time for APIs and <1s for interactive sites, but we couldn't agree about these in the discussion so left it that systems with latency alerting get to define the appropriate thresholds.

Force, for context, has p50 server-side latency of 1.25s and p99 of about 7s. The recent redesigns also shift a lot of work from the client to the server, so narrowly specifying server-side expectations might not capture important changes there.

However, I'd love for this to be more actionable somehow, such as maybe agreeing to standards for _new_ pages.

joeyAghion:
@eessex 2 open questions for you on this:
* How can we refine the item about "production environments being replicable locally?" I just want to make sure it's meaningful to developers and verifiable for each system.
* How readily does our existing code coverage tooling lend itself to enforcing that coverage only _increases_ in each PR?

playbooks/system-criticality.md
```diff
@@ -31,21 +31,23 @@ Retired systems, spikes, or time-bound experiments that aren't significantly use
...
+| Testing           | <ul><li>Automated tests</li><li>Tests are run on PRs and `master` by a Continuous Integration pipeline</li></ul>                                                                                                                     | <ul><li>&larr;ditto</li></ul>                                                            | <ul><li>&larr;ditto and...</li><li>Test coverage tooling</li></ul>                                                                                                     |
+| Deployment        | <ul><li>Zero-downtime deployment</li></ul>                                                                                                                                                                                           | <ul><li>&larr;ditto and...</li><li>High-fidelity staging environment</li></ul>           | <ul><li>&larr;ditto and...</li><li>Deployment and orchestration by Kubernetes</li></ul>                                                                                |
```
izakp: "Deployment and orchestration by use of Kubernetes rolling-updates"?

playbooks/system-criticality.md
```diff
@@ -31,21 +31,23 @@ Retired systems, spikes, or time-bound experiments that aren't significantly use
...
+| Performance       |                                                                                                                                                                                                                                      |                                                                                          | <ul><li>Latency-based monitors and alerting (tailored to service)</li></ul>                                                                                            |
```
izakp: Would add here at a minimum:
- Error rate
- P90 latency

Datadog prompts you to set these up when creating a new service

Optionally:
- % change in throughput
- P75 latency for more critical apps

eessex:
@joeyAghion in response to your questions - 
| How can we refine the item about "production environments being replicable locally?" I just want to make sure it's meaningful to developers and verifiable for each system.

This I think applies particularly to Webpack, and apps that have differing configurations for their production and development compilation processes. Generally it would mean that you can run the app with `NODE_ENV=production` and reproduce the production compilation setup locally.  Currently I believe Force is able to do this, though we do not have instructions for how to use the assets created with this (I think they would need to be hosted on S3). If this is overly complicated I think it is ok to remove, but it becomes a significant hurdle when changing webpack configurations.

| How readily does our existing code coverage tooling lend itself to enforcing that coverage only increases in each PR?
Codecov runs two checks on PR's, one to check ratios of coverage on the app, and another for the specific code diff. We can set this to fail if the overall coverage goes down, and also by an expected ratio (fail if less that 70% coverage). I'm seeing that the overall coverage ratio is a bit finicky in current repos (it always thinks it is changing up or down by 0.1%, even on package changes). I think this may be rooted in part with our multiple test runner setup (mocha/jest), and hope to see more clarity once this feature is added to a more compact app.

joeyAghion:
@eessex regarding code coverage changes, it sounds like you're saying we _can't_ currently enforce a "coverage doesn't go down" rule anywhere currently. Is that accurate? If this _is_ achievable, I'd be interested in trying it out on at least one project to gauge if it's a good starting point for these expectations.

eessex:
Yes we can definitely do a rule to fail if coverage goes down!

joeyAghion:
This framework has already been useful as the Platform team has worked on its related key result, so I'll merge. These follow-ups still exist:
* firm up the definition of production parity for the **development** category
* decide a response time for bounties/vulnerabilities for the **security** category

PR #254, opened 2019-10-07
iOS Learning Group Session Two Materials

ashfurrow:
This is a WIP because I have to finish a few sections at the end. The core parts are there, though, and I'm looking for ideas on:

- How we can get into the code a bit more (it's a bit theory-heavy right now).
- Possible homework questions.

Also, I would appreciate a second set of eyes on the parts where I talk about React vs React Native, to make sure I got the technical details right.


ashfurrow:
Probably easiest to review this with the rendered markdown preview: https://github.com/artsy/README/blob/ios-learning-group-sesh-two/resources/mobile/learning-group/session-two.md

resources/mobile/learning-group/session-two.md
```diff
@@ -0,0 +1,191 @@
...
+yarn start
+```
+
+## Why React Native?
+
+Artsy uses React Native because it
+[allows Artsy engineers working on the iOS and web front-ends to share the same idioms and tooling](https://artsy.github.io/blog/2019/03/17/three-years-of-react-native/).
+The primary benefit of using React Native is that
+[we get to use React](https://ashfurrow.com/blog/the-case-for-react-native/#react-itself-is-amazing), which is not
+only a better way of building iOS apps, but gives us those shared idioms and tools. (We're hoping to leverage React
```
xtina-starr: It might be worth looking into the code for one of the components in palette that has a web version and react native version. Maybe the button component?

resources/mobile/learning-group/session-two.md
```diff
@@ -0,0 +1,191 @@
...
+Native for Android in the longterm, but for now that's the motivation.)
+
+You can think of React Native as React, but instead of targeting `div`'s and the HTML DOM, you're targeting native
+views to whatever target platform you're running on. So, Artsy uses React Native to target the iOS `UIView`
+hierarchy. Developers write very similar, often identical components to the web, but at runtime, they get rendered
```
xtina-starr: This may be better place to dive into code using the example I mentioned above.

PR #265, opened 2019-11-01
Improves on-call responding instructions for updating the timeline

sweir27:
Thank you to @dblandin for doing a 👌 job at updating the timeline on a recent incident! (got to use that as an example. 😄 )


sweir27:
@izakp thanks for the feedback! Mind giving it another pass now that I've added more clarity around the "communications officer" role?

PR #289, opened 2020-02-04
Updates to migrate-redis and migrate-sidekiq playbooks

izakp:
Follow-up to https://github.com/artsy/docker-images/pull/62/

### DIFF
--- a/playbooks/migrating-redis.md
+++ b/playbooks/migrating-redis.md
@@ -12,19 +12,20 @@ Note: this migration process does not ensure data consistency if your app is act
 
 1) Check the old Redis URL
 
-Assuming your app is `$MY_APP`...
-
 ```
 hokusai [staging|production] get REDIS_URL
 ```
 
 This will be supplied to the migration container via `$SOURCE_REDIS_URL`
 
-2) Migrate all keys from the old Redis database to the new - plug in your app's name to $MY_APP the old and new Redis URLs to `$SOURCE_REDIS_URL` and `$DESTINATION_REDIS_URL`
+2) Migrate all keys from the old Redis database to the new - plug in the old and new Redis URLs to `$SOURCE_REDIS_URL` and `$DESTINATION_REDIS_URL`
 
-```
-kubectl --context [staging|production] run redis-migrate-$MY_APP --restart=Never --rm -i --tty --image artsy/redis-migrate --overrides='{"spec": {"containers": [{"tty": true, "stdin": true, "name": "redis-migrate-$MY_APP", "image": "artsy/redis-migrate:latest", "args": ["$SOURCE_REDIS_URL", "$DESTINATION_REDIS_URL"], "stdinOnce": true, "imagePullPolicy": "Always"}], "nodeSelector": {"tier": "background"}}, "apiVersion": "v1"}'
-```
+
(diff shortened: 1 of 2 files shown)


playbooks/migrating-redis.md
```diff
@@ -6,8 +6,38 @@ Unfortunately AWS' managed Redis deployments on Elasticache do not allow us to u
...
+```
+hokusai [staging|production] get REDIS_URL
+```
+
+This will be supplied to the migration container via `$SOURCE_REDIS_URL`
+
+2) Migrate all keys from the old Redis database to the new - plug in your app's name to $MY_APP the old and new Redis URLs to `$SOURCE_REDIS_URL` and `$DESTINATION_REDIS_URL`
+
+```
+kubectl --context [staging|production] run redis-migrate-$MY_APP --restart=Never --rm -i --tty --image artsy/redis-migrate --overrides='{"spec": {"containers": [{"tty": true, "stdin": true, "name": "redis-migrate-$MY_APP", "image": "artsy/redis-migrate:latest", "args": ["$SOURCE_REDIS_URL", "$DESTINATION_REDIS_URL"], "stdinOnce": true, "imagePullPolicy": "Always"}], "nodeSelector": {"tier": "background"}}, "apiVersion": "v1"}'
```
eessex: This command is different from the one that you've removed above, and the one we use as an example in the docker images repo instructions. Additionally, I think it is much less user-friendly in terms of readability. Can you elaborate on why `kubectl` is preferred here over `docker run`?

izakp: I suppose it's preferable as it requires no VPN connection, but yes agreed it's less user friendly.

Happy to converge on recommending we run these via `docker run` with a VPN connection rather than `kubectl run` in both cases.

Note that in order to use a VPN connection Docker has to be configured with:
```
"dns" : [
    "10.0.0.2",
    "8.8.8.8"
  ]
```

izakp: @joeyAghion @mzikherman wdyt?  Should we recommend `docker run` with a VPN connection or `kubectl run`?

izakp: Bump @joeyAghion @mzikherman

mzikherman: I would keep it simpler w/ VPN. I think it's sort of assumed you need VPN access to interact with our infrastructure (k8s stuff, etc.).

joeyAghion: If the command is simpler that way, I agree VPN is a reasonable thing to require. I'm not sure where those Docker DNS settings go though. If they're not obvious to set up, the more complicated kubectl command is acceptable too.

izakp: > not sure where those Docker DNS settings go though
https://lmgtfy.com/?q=docker+for+mac+config+file

izakp: Screenshot 2020-02-11 21 01 51

izakp: Sorry that was a bit passive aggressive - it's different on Docker for mac and Linux , where the config file lives in `/etc/docker` - my point is we can't spell out everything for everyone's local configuration - as an org we need to do better solving local config problems ourselves

eessex: We have docs for configuring this that could be linked back to here: https://www.notion.so/artsy/VPN-Configuration-60798c292185407687356997bf251d8c

izakp: Updated to use docker-run with a VPN connection and link to VPN setup docs in notion

PR #297, opened 2020-03-03
add playbooks/migrating-postgres-with-rubyrep.md

izakp:
Playbook for migrating Postgres databases with [Rubyrep](http://rubyrep.org/) using our [Docker image](https://github.com/artsy/docker-images/tree/master/rubyrep)

### DIFF
--- a/playbooks/migrating-postgres-with-rubyrep.md
+++ b/playbooks/migrating-postgres-with-rubyrep.md
@@ -62,7 +62,7 @@ Apply with `hokusai [staging|production] create --filename ./hokusai/export-old-
 
 Delete with `hokusai [staging|production] delete --filename ./hokusai/export-old-production.yml`
 
-To restore to the new datanase, create the folling Yaml file as `./hokusai/import-to-new-production.yml`
+To restore to the new database, create the folling Yaml file as `./hokusai/import-to-new-production.yml`
 
 ```
 ---
--- a/playbooks/migrating-postgres-with-rubyrep.md
+++ b/playbooks/migrating-postgres-with-rubyrep.md
@@ -1,0 +1,227 @@
+---
+title: Migrating Postgres with Rubyrep
+description: How to migrate a Postgres to a new instance using Rubyrep
+---
+
+# Migrating Postgres with Rubyrep
+
+To set up database replication via [rubyrep](rubyrep.org), do the following:
+
+1) Create the new Postgres database and user
+
+2) Start a shell in the application with the new database connection and load the app schema with:
+```
+hokusai [staging|production] run --env "DATABASE_URL=postgres://{new database credentials}" "bundle exec rake db:schema:load"
+```
+
+3) Check the database schemas.
+  3a) If you want to download a dump of the current db to a local postgres, you can check schemas by running `rake db:schema:dump` locally, and checking in the diff in `db/schema.rb`.  You can also inspect the `schema_migrations` tables across the databases and compare to migrations.
+  3b) Log into the new database and note all foreign key constraints.  You can get foreign key constraints for each table with `\d+ table`.  Save the constraints to later re-enable.
+
+4) Disable all foreign key constraints, for example for tables `foo` and `bar` you would run:
+```
+ALTER TABLE foo DROP CONSTRAINT fk_rails_fbc9d01ca0;
+ALTER TABLE bar DROP CONSTRAINT fk_rails_80eb82ccbf;
+```
+
+5) Create the file `hokusai/rubyrep-config.yml` (Credentials here are redacted but they should be supplied to Kubernetes and this file should not be checked in!) The "left" configuration refers to the source database and the "right" refers to the target database.  Note that for rails apps we can ignore both `ar_internal_metadata` and `schema_migrations` tables.
+
+```
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: myapp
+  name: rubyrep-myapp
+  namespace: default
+data:
+  default.conf: |
+    RR::Initializer::run do |config|
+      config.left = {
+        :adapter  => 'postgresql',
+        :database => 'REDACTED',
+        :username => 'REDACTED',
+        :password => 'REDACTED',
+        :host     => 'OLD_DATABASE',
+        :sslmode  => 'require',
+        :logger   => STDOUT
+      }
+
+      config.right = {
+        :adapter  => 'postgresql',
+        :database => 'REDACTED',
+        :username => 'REDACTED',
+        :password => 'REDACTED',
+        :host     => 'NEW_DATABASE',
+        :sslmode  => 'require',
+        :logger   => STDOUT
+      }
+
+      config.include_tables /./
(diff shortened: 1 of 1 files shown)


playbooks/migrating-postgres-with-rubyrep.md
```diff
@@ -0,0 +1,227 @@
...
+          configMap:
+            name: rubyrep-myapp
+```
+
+Create the deployment with:
+```
+$ hokusai [staging|production] create --filename ./hokusai/rubyrep.yml
+```
+
+7) When the database is synced and replication is active, we are ready to make a cutover. To avoid race conditions regarding auto-incrementing primary keys, we will have to accept a brief window of downtime while performing a cutover to the new database.
```
joeyAghion: Could we avoid the downtime and potential conflicts by increasing the id sequences of the new database (to sufficiently high values) in advance of the replication? Replication could continue to insert new data with the lower values, and any direct inserts after cut-over could pick up at the higher value.

izakp: I believe Rubyrep inserts rows _without_ making use of "autoincrement" so this may work if we are okay with skipping a few primary key ids.

@joeyAghion so you are suggesting that if, say `messagers_id_seq` is at `2000` we bump it to `2100` before cutover, allowing for any replicated messages to fill up to 2100 and any new messages to start from 2100?

A note on sequences - we are explicitly disabling Rubyrep's "adjust sequences" [config parameter](https://github.com/rubyrep/rubyrep/blob/master/lib/rubyrep/replication_initializer.rb#L63) as it will try and reorder sequences in both databases (i.e. left database would end up with only even primary keys and right database would end up with only odd) - note in the code the line "The sequence is always updated in both databases".

See http://www.rubyrep.org/configuration.html for sequence configuration options but I would hesitate to have rubyrep make any destructive changes to the source database.

joeyAghion: Yup that was my thought, but I don't know what's safest.

cavvia: As long as it's easy to configure the right ID offset for all tables this should work... we would have to choose an arbitrary offset but we can just set it cautiously high for all tables based on what we expect the highest throughput table to experience.

izakp: Thanks for the feedback @cavvia - since we don't have datadog monitoring set up I'll poll the production tables and report avg write throughput / minute here.

izakp: Impulse production seems to have a max throughput on the messages table of ~6 writes / minute

playbooks/migrating-postgres-with-rubyrep.md
```diff
@@ -0,0 +1,227 @@
...
+1) Create the new Postgres database and user
+
+2) Start a shell in the application with the new database connection and load the app schema with:
+```
+hokusai [staging|production] run --env "DATABASE_URL=postgres://{new database credentials}" "bundle exec rake db:schema:load"
+```
+
+3) Check the database schemas.
+  3a) If you want to download a dump of the current db to a local postgres, you can check schemas by running `rake db:schema:dump` locally, and checking in the diff in `db/schema.rb`.  You can also inspect the `schema_migrations` tables across the databases and compare to migrations.
+  3b) Log into the new database and note all foreign key constraints.  You can get foreign key constraints for each table with `\d+ table`.  Save the constraints to later re-enable.
```
izakp: One-liner to list all foreign key constraints for all tables in the `public` schema...
```
SELECT conrelid::regclass AS table_from
     , conname
     , pg_get_constraintdef(oid)
FROM   pg_constraint
WHERE  contype IN ('f')
AND    connamespace = 'public'::regnamespace  -- your schema here
ORDER  BY conrelid::regclass::text, contype DESC;
```

izakp:
Putting this on hold - I was able to reproduce the out-of-sync production impulse migration we saw yesterday @eessex @jpotts244 - it appeared to have missed 1000s of rows.

Apparently this tool deal well with small databases but not with large ones :/

As we spoke about in standup and as @cavvia suggested, will have to sacrifice downtime for data integrity, and will follow up with a PR in Impulse to utilize our `pg-data-sync` image to make a migration

izakp:
Removed #HOLD as this is unblocked

playbooks/migrating-postgres-with-rubyrep.md
```diff
@@ -0,0 +1,357 @@
...
+      dnsPolicy: ClusterFirst
+      restartPolicy: Never
+      terminationGracePeriodSeconds: 30
+```
+
+Apply with `hokusai [staging|production] create --filename ./hokusai/export-old-production.yml`
+
+Delete with `hokusai [staging|production] delete --filename ./hokusai/export-old-production.yml`
+
+To restore to the new datanase, create the folling Yaml file as `./hokusai/import-to-new-production.yml`
```
eessex: "datanase" should be "database"

playbooks/migrating-postgres-with-rubyrep.md
```diff
@@ -0,0 +1,375 @@
...
+```
+      table_from     |       conname       |                    pg_get_constraintdef
+--------------------+---------------------+-------------------------------------------------------------
+foo                  | fk_rails_fbc9d01ca0 | FOREIGN KEY (bar_id) REFERENCES bar(id)
+bar                  | fk_rails_80eb82ccbf | FOREIGN KEY (foo_id) REFERENCES foo(id) ON DELETE CASCADE
+```
+
+Save the constraints to later re-enable.
+
+4) Disable all foreign key constraints, for example for tables `foo` and `bar` you would run:
```
izakp: @eessex @joeyAghion @mzikherman thinking that this could be automated via a fancy PSQL script or (since my PSQL isn't so great) perhaps baked into RubyRep as a command, i.e. `rubyrep disable_fk_constraints`

playbooks/migrating-postgres-with-rubyrep.md
```diff
@@ -0,0 +1,375 @@
...
+  }' --image artsy/rubyrep
+  ```
+
+  The diff should show `0` in total for all tables.
+
+  7b) To avoid race conditions regarding auto-incrementing primary keys, reset _all_ table id sequences except for `rr_*` RubyRep tables and id sequences such as `rr_pending_changes_id_seq`, leaving an offset for writes to the new database and headroom for replicated inserts from the old.
+
+  Run this command for _all_ tables / table_id_sequences in the database.  To get a list of all tables and id sequences, log into the database via `psql` and run `\d+`
+
+  ```
```
izakp: Same here, perhaps a RubyRep command `rubyrep reset_id_sequences offset`?  My attempts to loop through all tables in PSQL while omitting all `rr_*` tables proved futile, but I think RubyRep already has biolderplat code to get an array of tables, excluding its own... it also used ActiveRecord so could then issue these composed statements 🤔

izakp:
Thanks for walking through this one more time and pointing out some of the ambiguity @eessex !  I think it's a sound method though we can clean it up a bit, it's quite verbose!

PR #301, opened 2020-03-19
Add instructions for OpsGenie alerts

eessex:
Follow up to https://artsyproduct.atlassian.net/browse/PLATFORM-2134, add instructions for using OpsGenie notifications/alerts.

### DIFF
--- a/playbooks/support/responding.md
+++ b/playbooks/support/responding.md
@@ -11,6 +11,14 @@ respond to these issues.
   These automated system alerts _may_ be the first sign of a problem but don't _necessarily_ warrant treatment as
   immediate incidents. E.g., load-balancers may route around failing EC2 instances or auto-scaling may bring up
   capacity in response to load.
+- Monitor automated uptime incidents created by OpsGenie.
+  - For high-criticality systems, OpsGenie will open incidents if
+    [Pingdom](https://my.pingdom.com/app/newchecks/checks) uptime checks are failing. Alerts will automatically
+    close if resolved.
+  - All engineers should have the [OpsGenie mobile app](http://itunes.apple.com/us/app/opsgenie/id528590328?mt=8)
+    installed and be logged in.
+  - Select your preferences for push notifications, text messages, and/or phone calls for alerting in your
+    [OpsGenie user notification settings](https://artsy.app.opsgenie.com/settings/user/notification).
 
 ## Step 2: Raise
 
@@ -76,9 +84,8 @@ respond to these issues.
 
 Discuss the incident's resolution in the dedicated Slack channel.
 
-- Start a video conference with Zoom to communicate in real-time with
-  other responders. Credentials to a shared Engineering Zoom Pro account are in
-  1Password.
+- Start a video conference with Zoom to communicate in real-time with other responders. Credentials to a shared
+  Engineering Zoom Pro account are in 1Password.
 - Fix if possible using [accumulated playbooks (wiki)](https://github.com/artsy/potential/wiki) 🔒.
 - If unable to fix using shared resources, contact the relevant [point-person](#point-people) for help and pair
   with them on a fix. Be sure to contribute lessons back to shared documentation. **You are not expected to know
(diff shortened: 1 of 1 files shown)


playbooks/support/responding.md
```diff
@@ -11,6 +11,14 @@ respond to these issues.
   These automated system alerts _may_ be the first sign of a problem but don't _necessarily_ warrant treatment as
   immediate incidents. E.g., load-balancers may route around failing EC2 instances or auto-scaling may bring up
   capacity in response to load.
+- Monitor automated uptime incidents created by OpsGenie.
+  - For high-criticality systems, OpsGenie will open incidents if
+    [Pingdom](https://my.pingdom.com/app/newchecks/checks) uptime checks are failing. Alerts will automatically
+    close if resolved.
+  - All engineers should have the [OpsGenie mobile app](http://itunes.apple.com/us/app/opsgenie/id528590328?mt=8)
```
izakp: Add Opsgenie app for Android https://play.google.com/store/apps/details?id=com.ifountain.opsgenie&hl=en

playbooks/support/responding.md
```diff
@@ -11,6 +11,14 @@ respond to these issues.
...
+    installed and be logged in.
+  - Select your preferences for push notifications, text messages, and/or phone calls for alerting in your
```
izakp: Perhaps recommend:
- Alert after 1 minute via push notification
- Alert after 5 minutes to SMS
- Alert after 20 minute via phone call

playbooks/support/responding.md
```diff
@@ -92,22 +99,31 @@ change should be addressed by the relevant team.
...
+alerting external stakeholders. In general, you should post to the timeline as any progress is made on the
+incident, and much more frequently than you post to the StatusPage.
 
-Keep in mind that the timeline will be referenced later while reviewing the incident. It's important to have as much information captured as possible for piecing together the steps to resolution. If you forget to add an entry in the moment, no worries! OpsGenie allows you to add entries after-the-fact with the correct timestamp.
+Keep in mind that the timeline will be referenced later while reviewing the incident. It's important to have as
+much information captured as possible for piecing together the steps to resolution. If you forget to add an entry
+in the moment, no worries! OpsGenie allows you to add entries after-the-fact with the correct timestamp.
 
-For an example of a comprehensive timeline, check out [this postmortem report](https://artsy.app.opsgenie.com/reports/post-mortem/b5710803-4bcb-4f0b-a906-4a2343521328/detail)🔒 from 10/29/19.
+For an example of a comprehensive timeline, check out
```
izakp: 👍

playbooks/support/responding.md
```diff
@@ -11,6 +11,14 @@ respond to these issues.
...
   immediate incidents. E.g., load-balancers may route around failing EC2 instances or auto-scaling may bring up
   capacity in response to load.
+- Monitor automated uptime incidents created by OpsGenie.
+  - For high-criticality systems, OpsGenie will open incidents if
+    [Pingdom](https://my.pingdom.com/app/newchecks/checks) uptime checks are failing. Alerts will automatically
+    close if resolved.
+  - All engineers should have the [OpsGenie mobile app](http://itunes.apple.com/us/app/opsgenie/id528590328?mt=8)
+    installed and be logged in.
+  - Select your preferences for push notifications, text messages, and/or phone calls for alerting in your
+    [OpsGenie user notification settings](https://artsy.app.opsgenie.com/settings/user/notification).
```
izakp: Perhaps a note about acknowledging ( acking ) the alert - we want to get folks in this habit before we define clear response SLAs and add escalation policies

eessex:
Comments have been addressed, this is ready for review again.

playbooks/support/responding.md
```diff
@@ -11,6 +11,23 @@ respond to these issues.
   These automated system alerts _may_ be the first sign of a problem but don't _necessarily_ warrant treatment as
   immediate incidents. E.g., load-balancers may route around failing EC2 instances or auto-scaling may bring up
   capacity in response to load.
+- Monitor and acknowledge automated uptime incidents created by OpsGenie.
+
+  - For high-criticality systems, OpsGenie will open incidents if
```
dblandin: Does OpsGenie open an incident automatically or does it create an alert in response to failing uptime checks from Pingdom? I think it might be the latter.

Once acknowledged, if the responding on-call engineers decide to open an incident in response, there are ways within the OpsGenie UI to associate open alerts with the new incident. Might be worthwhile to explore and document as a follow-up.

playbooks/support/responding.md
```diff
@@ -11,6 +11,23 @@ respond to these issues.
...
+    [Pingdom](https://my.pingdom.com/app/newchecks/checks) uptime checks are failing. Alerts will automatically
+    close if resolved.
```
dblandin: What do you think of clarifying this behavior to something like:

Alerts will automatically close within OpsGenie if resolved via Pingdom.

playbooks/support/responding.md
```diff
@@ -11,6 +11,23 @@ respond to these issues.
...
+    [Android](https://play.google.com/store/apps/details?id=com.ifountain.opsgenie&hl=en) installed and be logged
+    in.
+  - Select your preferences for push notifications, text messages, and/or phone calls for alerting in your
+    [OpsGenie user notification settings](https://artsy.app.opsgenie.com/settings/user/notification).
+
+    Recommended settings:
+
+    - Immediately via push notification
+    - After 5 minutes via SMS
+    - After 15 minutes via phone call
```
dblandin: Currently, I don't think responders have the ability to customize these notification preferences. We're setting that up globally via a Central Notification Template:

image

https://docs.opsgenie.com/docs/central-notification-templates
https://artsy.app.opsgenie.com/settings/notificationSettings/detail/5ebe8a41-9752-4b75-8bd6-a16d52824759

I think the reason for setting this up originally was to ease on-boarding. New on-call members wouldn't have to set up their own notification preferences. But if we'd rather require on-call responders to set these up before shifts, then we could certainly consider removing the global override. There's also the question of whether having consistent notification rules across the team is more valuable than any single member being able to customize each notifcation methods and thresholds.

eessex: Added this at the suggestion of @izakp but I'm happy to just let everyone use the defaults.

PR #306, opened 2020-05-01
Add Heroku as a hold

zephraph:
It was briefly [discussed](https://artsy.slack.com/archives/CP9P4KR35/p1588197242487700?thread_ts=1588196125.480000&cid=CP9P4KR35) that heroku should be added to a `hold` state.

### DIFF
--- a/playbooks/technology_radar/artsy-tech-radar.csv
+++ b/playbooks/technology_radar/artsy-tech-radar.csv
@@ -60,3 +60,4 @@ MongoDB,hold,tools,FALSE,
 Pingdom,hold,tools,TRUE,
 Semaphore,hold,tools,FALSE,
 Travis CI,hold,tools,TRUE,
+Heroku,hold,platforms,FALSE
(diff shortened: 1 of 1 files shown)


playbooks/technology_radar/artsy-tech-radar.csv
```diff
@@ -59,4 +59,5 @@ Kafka,hold,tools,FALSE,
 MongoDB,hold,tools,FALSE,
 Pingdom,hold,tools,TRUE,
 Semaphore,hold,tools,FALSE,
-Travis CI,hold,tools,TRUE,
\ No newline at end of file
+Travis CI,hold,tools,TRUE,
+Heroku,hold,platforms,FALSE
```
joeyAghion, line 63: A short explanation could help here. Borrowing from the kubernetes line above, something like: "We prefer kubernetes where we have shared solutions for authorization, logging, monitoring, alerting, scaling, and routing."

zephraph:
```suggestion
Heroku,hold,platforms,FALSE,"We prefer kubernetes where we have shared solutions for authorization, logging, monitoring, alerting, scaling, and routing."
```

PR #315, opened 2020-06-25
fix: update values

adamvert:
Insert the new values, copied from the [slide deck](https://docs.google.com/presentation/d/1Hk8wX9SRmKlf9mQsYFsGVF7jbHOFikaeFRVo7PEG0DE/edit#slide=id.g7f6d69116f_6_74). There might be more stuff here that's out of date, but this is what jumped out at me on a first reading.

@iskounen I assigned this to you, since it looks like you've been active on other open PRs. I guess that this repo could do with some ❤️ generally, but I can see how it tends to drop below the radar! I'm happy to do some more updating as I discover things, if that would make sense, but I don't really feel like I'm qualified to make the kind of judgement calls that would be needed to bring some of it up to date.

### DIFF
--- a/culture/what-is-artsy.md
+++ b/culture/what-is-artsy.md
@@ -22,57 +22,37 @@ Artsy?" Well, Artsy pays our wages, so we probably owe you and the company an ex
 
 ## Artsy Values
 
-Artsy has 5 core values:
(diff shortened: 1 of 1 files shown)


joeyAghion:
Thanks!

PR #321, opened 2020-07-29
Remove redundant section on hokusai configure.

artsyjian:
Update to Hokusai doc.

- Made `Configure Hokusai` a higher level heading.
- Removed `Configuring access to Kubernetes` section. It's content is already covered by `Configure Hokusai`.


izakp:
👍

PR #326, opened 2020-08-14
fix: clarify on-call procedure

pvinis:
i missed it, so maybe others might too 😬.


playbooks/support/README.md
```diff
@@ -6,6 +6,7 @@
 - [Wait!](#wait)
 - [Process Overview](#process-overview)
 - [On-Call Responsibilities](#on-call-responsibilities)
+- [Setting up for your on-call](#setting-up-for-your-on-call)
```
sweir27, line 9: #minor Can we call this `Preparing for your on-call shift` instead?

admbtlr (approved):

PR #328, opened 2020-09-08
Make About Artsy links in README absolute

starsirius:
We've been including the "About Artsy" section in README in public repos. However, some links are _relative_ so they break in other repos. This turns them into absolute links so copying and pasting still works.


jonallured:
Thank you!! ❤️ ❤️ ❤️

PR #329, opened 2020-09-16
Update deployments playbook to reflect more consistent process and mature tools

joeyAghion:
This doc was last updated during the early part of our transition to more consistent CI and release tooling. Happy to say we can drop most of it finally, and instead point readers (new engineers, especially) to the templates and tools that will mostly do the right, standard thing.

Depends on doc updates pending in https://github.com/artsy/horizon/pull/221.

### DIFF
--- a/playbooks/deployments.md
+++ b/playbooks/deployments.md
@@ -26,142 +26,64 @@ description: How systems are deployed at Artsy
 
 _\*Not all of these goals are routinely achieved._
 
-_[Jump ahead to **recommendations**.](#recommendations)_
-
-## Current state
-
-- A few different approaches, largely historical or based on local teams' preferences.
-- In all cases, CI deploys to staging
-- In almost all cases, **CircleCI** is responsible for CI
-- Want to deploy? Check the **README** of any system to be sure.
-
-### Hosting Environments:
-
-- Heroku (Radiation, Impulse, Positron, ...)
-- AWS OpsWorks (Gemini, Causality)
-- **Kubernetes** (Gravity, Force, Metaphysics, Volt, Diffusion, Doppler, ...)
-
-### Approaches:
-
-#### Run a single command from a developer's check-out
-
-E.g.:
-
-- `hokusai pipeline promote` (Metaphysics, Diffusion, Vibrations...)
-- `git push heroku master` (...?)
-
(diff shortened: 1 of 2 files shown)


playbooks/deployments.md
```diff
@@ -26,142 +26,64 @@ description: How systems are deployed at Artsy
...
+    file.
+  - Check the _Allow write access_ box and click the _Add key_ button to save the new key.
+  - In the CircleCI project settings, go to _SSH Permissions_ > _Add SSH Key_.
+  - Enter `github.com` for _Hostname_ and the contents of the _private_ key file for _Private Key_, then click _Add
+    SSH Key_ to save.
+- In the projects' settings, generally speaking:
+  - _build forked pull requests_ should be enabled
+  - _pass secrets to builds from forked pull requests_ should be disabled
+
+## "Hot" fixes and roll-backs
```
izakp, line 82: This is great 🎉

playbooks/deployments.md
```diff
@@ -26,142 +26,64 @@ description: How systems are deployed at Artsy
...
+
+Sometimes there are urgent fixes to release. We always prefer using the typical pipeline to "roll forward," even in
+cases of reverted PRs or timely fixes. When it's critical to avoid the delay of a full CI run or staging deploy,
+it's possible to simply `hokusai production deploy <tag>` where _tag_ is an image tag (including git SHAs) and can
+refer to a previous release such as `production--2018-09-25--10-19-2`. (Run `hokusai registry images` to see recent
+tags.) Projects that depend on other release operations (e.g., to compile assets or update configuration) may
+require additional steps.
```
izakp, line 89: Thinking good to stress here that this will deploy the docker image for the project (i.e. the code and dependencies as an artifact) but will _not_ include orchestration (changes to Kubernetes yaml files) as well as any release operations.

Somewhat tangential @joeyAghion do we want to deprecate `hokusai [staging|production] update`?

joeyAghion: That's a good point [that I didn't realize] about the orchestration.

I've used the `... update` command recently, either with `--dry-run` to validate some in progress changes, or during incidents. From my perspective it would be good to keep, even as an "advanced" command that most folks don't end up needing to run.

PR #330, opened 2020-09-17
Updating stand up agenda

SamRozen:
- Adding explicit call out to async communication
- Adding new section for optional Q&A / open discussion


erikdstock:
👍 from me, both of these would be super helpful.

Issue #309, opened 2020-05-07
RFC: Automate or abandon forks


zephraph:
## Proposal:

Completely abandon fork workflows in favor of branches 

OR

Automate privileged forks to reduce friction in those workflows

## Reasoning

We've had a long, colorful history with forks. We've got a [smattering of projects that disallow fork PRs via peril](https://github.com/artsy/peril-settings/blob/fbcfe31642526ea806f887a600707526c652623b/org/ossPRsForbidForks.ts#L7). We have had (still have?) github tokens for commenting exposed to the public _intentionally_ because danger ran some important tasks that we wanted to make sure there were comments posted on the PR for. 

The friction is that we can't (currently) share privileged information (tokens) with forks and thus they break a lot of tooling/automation. This _is_ something we can surmount technically though. We could have automation that listens for new opened fork PRs and automatically creates a tag or branch that's pushed upstream. With certain naming conventions, either of those strategies could directly report status directly back to the fork branch. It'll take some iteration to get right, but this is automation we could do once to allow folks to work however they want.

We could also just stop using forks entirely. That'll change how some people work, but it'll completely eliminate the need for any special tooling. 

So those are (as I see it) our options. I'd welcome other options if I missed them. 

## How is this RFC resolved?

We decide on an approach and take action towards it.

jonallured:
> and thus they break a lot of tooling/automation

I could use some further explanation on this point - what breaks?

zephraph:
Anything that needs tokens or credentials to run. Deploying canaries, commenting on PRs, doing stuff in slack, etc.

dblandin:
@zephraph Thanks for this RFC :100: 

I'm in favor of Artsy GitHub organization members using branches within the organization's repositories instead of forks.

For external contributors to our open-source projects, I think any privileged tasks could skip or degrade so that the core function of building and running tests continues for fork PRs.

A nice-to-have would be to post some message if we detect a PR running from a fork _and_ the author is in the Artsy GitHub organization. It's difficult to roll out process changes like this. A little nudge when appropriate and acceptanting gradual adoption might be easier than a more aggressive mandate.

ashfurrow:
> We have had (still have?) github tokens for commenting exposed to the public _intentionally_ ... 

[...]

PR #331, opened 2020-10-01
fix: change from forking to branching on main repo

admbtlr:
Following an [RFC](https://github.com/artsy/README/issues/309) and much discussion, we are now working in branches on the main repo instead of forks.

### DIFF
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -9,7 +9,7 @@ description: How we work together
 - [Working together](#working-together)
   - [Project Management](#project-management)
   - [Workflow](#workflow)
-    - [Fork-and-branch](#fork-and-branch)
+    - [Working in branches](#working-in-branches)
     - [Commits](#commits)
     - [Pull requests](#pull-requests)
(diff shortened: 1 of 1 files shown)


ashfurrow (commented):
I mean, _I_ think this is a good idea, but should this be discussed in an RFC? Historically this has been a contentious topic at Artsy, but I could be missing recent discussions or context.

ashfurrow (approved):
Ahhhhh just saw the link, cool!!

admbtlr:
Woah, that was quick! I thought this was already decided, after the RFC I linked to?

admbtlr:
Is that the right format for the label? 🤔

ashfurrow:
Hmm, let's see. #squashongreen

ashfurrow:
Hmm that should have done it. I've deleted the label so Peril will re-create it: #squashongreen

ashfurrow:
There we go, weird!

ashfurrow:
Oh it might not squash if the label was added after the success event from CI, I'd just squash manually 👍

PR #332, opened 2020-10-02
feat: add a 1:1:1 playbook

admbtlr:
I've been meaning to do this for a while. There aren't really any playbooks on management topics, but it seems like this is a good thing thing to write about, and a good place to put it.


playbooks/one-on-one-on-ones.md
```diff
@@ -0,0 +1,84 @@
...
+> 2. <Individual>, this will also be an opportunity for you to agree/disagree with how these things are
+>    described/characterized, and also an opportunity for you to have a complete picture of the info shared between
+>    <Old Manager> and me. The goal is to make sure we're transparent in the handoff of your career, how you want
+>    to grow, and the best ways I can support you.
+>
+> This stuff can naturally be very awkward for all involved :) This is okay (and normal!). The payoff for this
+> awkwardness is clarity, a smooth transition, and enabling me to fully support <Individual>'s career path.
+>
+> I hope that you're both happy and comfortable with this arrangement. If not, or if you're unsure about anything,
+> feel free to let me know your questions, doubts or complaints!
```
joeyAghion, line 70: (Some funky formatting in the rendered view of this section...)

joeyAghion (approved):
Great contribution!

Question to consider, that I wondered about myself in participating in 1-on-1-on-1s: should the "artifacts" of management relationships (like written reviews, feedback, or goals) be handed off as part of this process?

SamRozen:
Great topic to cover. Thanks for writing this up and then, even if it's not a blog post, should we tweet this from ArtsyOpenSource account?

pvinis (approved):
As a person that went through this process, this makes sense. 👍

starsirius (approved):
Thank you for documenting this and making it easy to follow!

admbtlr:
> Question to consider, that I wondered about myself in participating in 1-on-1-on-1s: should the "artifacts" of management relationships (like written reviews, feedback, or goals) be handed off as part of this process?

That's a good question. Maybe it depends on whether the written artefacts have been shared between the manager and report, or whether they are just the manager's notes? In the former case then I would say yes, in the latter no.

iskounen (approved):
This is wonderful documentation! Thank you for recording this process.

sweir27 (approved):
This is great! Thanks for the writeup.

Organization question: Should this be nested under a "Management Resources" or "Management Playbooks" section?

Re: Artifacts, I agree about not sharing 1:1 notes. I believe that reviews and feedback in Lattice are automatically transferred to the new manager, which in my experience has been sufficient! Might be worth calling out here (but no strong opinion on that front).

admbtlr:
THat's an interesting idea - we have playbook subdirs for `agile`, `support` and `technology_radar`, so maybe we could add one for `management`?

admbtlr:
Does this look OK to everyone?

joeyAghion (approved):
I see both docs still, but other than that 👍 .

admbtlr:
How do we go about tweeting this from ArtsyOpenSource account?

PR #338, opened 2020-10-17
small docs update while running through that list

artsyit:
That step was needed.


joeyAghion (commented):
In my experience the fingerprint isn't necessary, maybe because [by default _all_ configured keys are added](https://circleci.com/docs/2.0/configuration-reference/#add_ssh_keys)?

pvinis:
lol accidental PR from artsyit account instead of mine but ok.

I saw it should include all, but it kept complaining until I added it, so I didn't investigate further. 🤷

SamRozen:
@pvinis @joeyAghion bumping this up since it's been open for a while. Is it still relevant?

joeyAghion:
I suspect the difference (between some commands needing this step and others that don't) has to do with whether pre-built steps/orbs are being run or ad hoc commands using the newly built application container. I committed a small tweak to the language to reflect that, and will merge.

PR #339, opened 2020-10-23
[RFC] Recommend naming working branches like <github_user_name>/<topic>

joeyAghion:
Our shared repositories have started accumulating crufty branches since https://github.com/artsy/README/issues/309. Automatically deleting merged branches helps, but there's still the potential to accumulate branches from works-in-progress or abandoned work streams, and even the potential to conflict on common names.

Let's recommend the convention of `<github_user_name>/<topic>`. This is the nicest pattern I've seen, among others like `<user>-<topic-...>` or `<jira#>-...`.

### DIFF
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -35,11 +35,10 @@ README if you want to know who to start asking questions to.
 
 ### Working in branches
 
-After using the [fork-and-branch](https://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/) workflow
-for several years, Artsy teams now prefer to work in branches directly on the main repository. This means that
-developers should clone the repo, create a branch from `master` for the work they are doing, and then push that
-branch back to the main repo. [Pull requests](#pull-requests) are then made from the working branch back to the
-`master` branch.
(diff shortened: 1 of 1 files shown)


zephraph:
I'm 👍 generally, feels fine if folks fine it helpful. Would be nice to help folks adhere by adding a peril rule to warn and figure out if there's any local automation (git hooks?) that helps adhere to the pattern without a lot of extra thought. 

Let's make it easy.

damassi:
Related to the branch naming in order to reduce clashes, when @zephraph just went through and generously cleaned up all of our old stale branches (100+) it blew up a) notifications; and b) ability to see keep track of whats relevant since the changes swamped things actually needing review. For the sake of tidiness this is totally fine, but it would be great if we could solve this somehow and prompt devs to clean up branches after merges or automatically delete after a merge on green.

zephraph:
I just added the flag in force to auto delete branches after they're merged. `staging` and `release` are protected now so they won't be deleted. Should solve that problem.

dblandin (approved):
Sounds good to me 👍

joeyAghion:
Resolved! Going forward, please name branches on shared repos like `<github_user_name>/<topic>`. Please also make an effort to clean up any legacy-named branches when convenient.

PR #341, opened 2020-10-30
fix: update values again

admbtlr:
we also need to update the _How do we make money_ section


ashfurrow (approved):
Good catch – I didn’t even know we had the draft values on GitHub. I have a goal next quarter to lead a review of our Engineering Guiding Principles, with the context of these new values. I’m open to pairing in that!

admbtlr:
Sounds good! I'd definitely be interested in pairing on that.

PR #342, opened 2020-11-03
RFC: Adds suggestions for Code Reviewing

sweir27:
## Context
Teams (and individuals) have been experimenting with adding more clarity around their PR comments. Most recently, this manifested in some teams trying out [Conventional Comments](https://conventionalcomments.org) guidelines.

However, it can be jarring or confusing to see conventional comments-style code review labels without explanation. See discussion (and the impetus for this RFC!) in [this slack thread](https://artsy.slack.com/archives/C02BC3HEJ/p1598380026075800).

While I don't believe we need to require code review comments to follow any particular standard format at this time, I do think it would be helpful to reference Conventional Comments and provide some suggestions for quality code reviews.

If we find that code review comments are consistently low-quality or causing confusion, we can certainly revisit how codified this process should be on the team!

## Proposal
Merge this PR, which provides _some_ guidance for code review comments, and a section that can be further expanded in the future.

### DIFF
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -12,6 +12,7 @@ description: How we work together
     - [Fork-and-branch](#fork-and-branch)
     - [Commits](#commits)
     - [Pull requests](#pull-requests)
+    - [Code Reviews](#code-reviews)
     - [Testing](#testing)
     - [Deployment](#deployment)
   - [Continuous improvement](#continuous-improvement)
@@ -141,6 +142,15 @@ Further reading:
 - [On empathy and pull requests](https://slack.engineering/on-empathy-pull-requests-979e4257d158)
 - [Rules of communicating at GitHub](https://ben.balter.com/2014/11/06/rules-of-communicating-at-github/)
 - [RFC on defining PR roles at Artsy](https://github.com/artsy/README/issues/177)
+
+### Code Reviews
+
+The best PR comments are clear in their intention. While we don't mandate that PR comments follow any specific
+format, some people have found it helpful to adopt the standards outlined in
+[Conventional Comments](https://conventionalcomments.org).
+
+In particular, we encourage the use of `blocking` and `non-blocking` to signal whether or not a question or comment
+should be resolved before merging.
 
 ### Testing
 
--- a/playbooks/support/README.md
+++ b/playbooks/support/README.md
@@ -40,12 +40,12 @@ sprints. See [below](#examples) for some common examples.
 ## Wait!
 
 If you are new to on-call or looking for a refresher, we recommend watching the
-[on-call onboarding session](https://drive.google.com/file/d/1nGtNLCwP9zOBxhocnAFI-Ua1xs7eJYyD/view?usp=sharing) 🔒 we
-ran on June 13, 2019 which talks through your responsibilities as an on-call engineer.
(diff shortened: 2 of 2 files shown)


damassi:
I've found it pretty helpful to use the `(nonblocking)` or `(blocking)` label for emphasis, but the other forms have never felt (imo) that they've added much clarity to a discussion. Curious what other folks think.

playbooks/engineer-workflow.md
```diff
@@ -142,6 +143,14 @@ Further reading:
 - [Rules of communicating at GitHub](https://ben.balter.com/2014/11/06/rules-of-communicating-at-github/)
 - [RFC on defining PR roles at Artsy](https://github.com/artsy/README/issues/177)
 
+### Giving Code Reviews
+
+The best PR comments are clear in their intention. While we don't mandate that PR comments follow any specific
+format, some people have found it helpful to adopt the standards outlined in
+[Conventional Comments](https://conventionalcomments.org).
+
+In particular, it helps with clarity and velocity to specify when a given comment is `blocking` or `non-blocking`.
```
sweir27, line 152: @damassi I agree w/ you re: blocking and non-blocking! Tried to call that out here but let me know if it could be stronger.

damassi: This looks 👍 to me

playbooks/engineer-workflow.md
```diff
@@ -12,6 +12,7 @@ description: How we work together
     - [Fork-and-branch](#fork-and-branch)
     - [Commits](#commits)
     - [Pull requests](#pull-requests)
+    - [Giving Code Reviews](#giving-code-reviews)
```
joeyAghion, line 15: Could just be "Code review."

joeyAghion (approved):

playbooks/engineer-workflow.md
williardx: If the goal is to get people to do this then I think this language should be stronger, e.g. "In particular, we encourage the use of `blocking` and `non-blocking` to signal whether or not a question or comment should be resolved before merging."

pepopowitz (approved):
❤️ ❤️ ❤️

playbooks/engineer-workflow.md
pvinis: "Code reviews" :D

pvinis (approved):
I like this 👍

anandaroop (approved):

damassi (approved):

playbooks/engineer-workflow.md
```diff
@@ -142,6 +143,15 @@ Further reading:
...
 - [RFC on defining PR roles at Artsy](https://github.com/artsy/README/issues/177)
 
+### Code Reviews
+
+The best PR comments are clear in their intention. While we don't mandate that PR comments follow any specific
+format, some people have found it helpful to adopt the standards outlined in
+[Conventional Comments](https://conventionalcomments.org).
+
+In particular, we encourage the use of `blocking` and `non-blocking` to signal whether or not a question or comment
+should be resolved before merging.
```
starsirius, line 153: 👍  This could work well with Github's "Request changes" feature if we want to make blocking comments more obvious.

starsirius (approved):

sweir27:
Updated based on the comments, but this has received lots of support so I'm going to merge it in! 🚀 

PRs to update always welcome. 😄

PR #347, opened 2020-11-20
Replace Dropbox links with Google Drive links on Lunch & Learn highlights page

annacarey:
Since we've migrated over to Google Drive, the Dropbox links were broken. I'm pretty sure the links I used would work for anyone at Artsy but we should double check. (Could also consider adding a few more highlights from the archive in a future PR since there are so many good ones!)

### DIFF
--- a/resources/lnl.md
+++ b/resources/lnl.md
@@ -3,28 +3,31 @@ title: Highlights from our L&Ls
 description: The best-of from our archives
 ---
 
+Recordings of past lunch and learns can be found in the Shared Google Drive under
+[__Everyone at Artsy/PDDE/Engineering/Lunch & Learn](https://drive.google.com/drive/u/0/folders/1X8w7iFbdeVwi6v_xWLWfQdeJ81iewYWB)
+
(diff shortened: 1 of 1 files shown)


dblandin:
Confirming that these links work for me! `/drive/u/0/` will refer to your primary Google account vs`/drive/u/1` or `/drive/u/2` but it's easy to switch to the right account when prompted:

Screen Shot 2020-11-20 at 11 47 15 AM

annacarey:
@dblandin would `/drive/u/1` or `/drive/u/2` be better then? Or does it not matter b/c `0`, `1`, and `2` may or may not point at `@artsymail` depending on the person?

dblandin:
> Or does it not matter b/c 0, 1, and 2 may or may not point at @artsymail depending on the person?

Exactly. Which account maps to 0/1/2 entirely depends on the user so `/drive/u/0` sounds good to me. 👍

dblandin (approved):

starsirius:
Thank you for fixing them! 💪

PR #348, opened 2020-11-30
RFC: Add sections explicitly outlining RFC implementation and approval to the RFC template and documentation

williardx:
## Proposal

Update the RFC docs and templates to include an explicit section on ~(1) how an RFC will be implemented and (2)~ how an RFC is resolved. This section should be a precise description of how we'll know when the proposal has been accepted.

## Reasoning

The goal of this RFC is to drive more clarity about the implementation and resolution of RFCs. Clearer guidelines around implementation and resolution will allow team members (especially newer members) to contribute more easily to the RFC process.

~While in practice we do include implementation plans for RFCs adding this section to the template makes it explicit and serves as a reminder to think about how to roll out the proposal. This is especially important for changes that have large scope and might require broader feedback and alignment to fully implement.~

Our current RFC resolution process is fuzzy and does not address complex changes well. To help teammates all be on the same page and to create alignment around change more easily, we should be explicit about how an RFC will be considered approved when we create an RFC. If _how_ to approve a given RFC creates some conflict, that's good! That means we'll have a chance as a team to work out issues and create stronger consensus for a proposal.

## Exceptions

We should strive for impact over perfection. RFC creators should use their best judgment about how to use these sections. If the change is minor then it might not require a detailed implementation or resolution section.

## How this RFC is implemented

This RFC updates the RFC template and docs and will be implemented as soon as this PR is merged.

## How this RFC is resolved

This RFC will be considered approved if (1) @sweir27 and @joeyAghion explicitly approve as primary stakeholders for the RFC process and (2) if the level of support from the engineering team at large is at least `Majority acceptance, with conflicting feedback.` However, open to other ways to consider this RFC "approved" and think it makes for a good discussion!

### DIFF
--- a/playbooks/rfcs.md
+++ b/playbooks/rfcs.md
@@ -43,19 +43,14 @@ Create a plan for how the RFC will be implemented. If the change is complex or l
 feedback process to iterate on the proposal as you implement and creating milestones and/or metrics to track
 progress.
 
-Decide who needs to know about this proposal, what role teammates will play in the process and how the RFC will be
-considered approved.
+Decide how you'll resolve the RFC:
 
-- What is the scope of the proposal? Does the whole engineering team need to know or just a smaller group of
-  engineers?
-- Who should actively approve of this RFC? Who should give feedback?
-- Does the change make sense without further discussion from engineering leadership?
-
-Here are some examples of how an RFC could be resolved:
-
-- X% of the engineering team actively approves this change.
-- The point people of the impacted repos approve this change.
-- Engineering leadership approves the changes after collecting feedback from the team.
+- Specify whose feedback is necessary on this RFC with review requests.
+- Specify for how long feedback will be collected and incorporated into the proposal. At least a week for feedback
+  is appropriate for any proposal impacting all of engineering.
+- Specify how approval will ultimately be decided. Does this depend on X% of the engineering team actively
+  approving, point people of certain repos, or perhaps engineering leadership? If specific people need to approve
+  use an `@` mention in the RFC to say so.
 
 Create an issue, and work from this template:
 
@@ -64,6 +59,11 @@ Create an issue, and work from this template:
     ## Proposal:
 
     Apply a spell checker to every markdown document that appears in a PR.
+
+    As a trial, we'll add a new Peril rule that implements spellchecking on every PR that includes
+    a Markdown document in README. We'll then gather asynchronous feedback from engineers about the
+    new Peril rule and how it impacted their workflow. We'll confirm we want this rule and incorporate
+    any further changes to the original RFC. We'll then roll out the Peril rule to all systems.
 
     ## Reasoning
 
@@ -83,16 +83,11 @@ Create an issue, and work from this template:
 
     You can see our discussion [in slack here](/link/to/slack.com)
 
-    ## How this RFC is implemented
-
-    As a trial, we'll add a new Peril rule that implements spellchecking on every PR that includes
-    a Markdown document in README. We'll then gather asynchronous feedback from engineers about the
-    new Peril rule and how it impacted their workflow. We'll confirm we want this rule and incorporate
-    any further changes to the original RFC. We'll then roll out the Peril rule to all systems.
-
     ## How this RFC is resolved
 
-    This RFC will be considered approved when 50% of the engineers at Artsy actively approve this proposal.
+    We'll collect feedback on this RFC from the team for a week. We'll consider this RFC approved
+    if 50% of the engineering team actively approves of this change. Add a 👍 or 👎  reaction to
+    this proposal to vote.
 
 Give a week for discussion.
 
@@ -124,10 +119,6 @@ If possible, please use one of these for the Level of Support section:
 - `5: Unclear Resolution.`
 - `6: RFC Rejected.`
 - `7: RFC Rejected, with Conflicting Feedback.`
-
-Resolving an RFC requires you to have some nuance about the feedback. If it seems to be unresolved, or still active
-a week later, then calling a town hall style meeting for people involved will probably shake out some kind of
-resolution.
 
 [potential]: https://github.com/artsy/potential/
 [mobile]: https://github.com/artsy/mobile
--- a/playbooks/rfcs.md
+++ b/playbooks/rfcs.md
@@ -48,9 +48,11 @@ Decide how you'll resolve the RFC:
 - Specify whose feedback is necessary on this RFC with review requests.
 - Specify for how long feedback will be collected and incorporated into the proposal. At least a week for feedback
   is appropriate for any proposal impacting all of engineering.
-- Specify how approval will ultimately be decided. Does this depend on X% of the engineering team actively
-  approving, point people of certain repos, or perhaps engineering leadership? If specific people need to approve
-  use an `@` mention in the RFC to say so.
+- Specify how approval will ultimately be decided. By default most proposals can be considered approved if there is
+  no objection. However, if the change is complex or has wide impact you may want a more structured approval
+  process. For example, does this depend on X% of the engineering team actively approving, point people of certain
+  repos, or perhaps engineering leadership? If specific people need to approve use an `@` mention in the RFC to say
+  so.
 
 Create an issue, and work from this template:
 
--- a/playbooks/rfcs.md
+++ b/playbooks/rfcs.md
@@ -39,6 +39,24 @@ Find the right repo:
 - Does it affect the whole dev team? (use [README](https://github.com/artsy/README))
 - Can't figure it out or needs to be private? (use [Potential][] as a fallback)
 
+Create a plan for how the RFC will be implemented. If the change is complex or large in scope consider defining a
+feedback process to iterate on the proposal as you implement and creating milestones and/or metrics to track
+progress.
+
+Decide who needs to know about this proposal, what role teammates will play in the process and how the RFC will be
+considered approved.
+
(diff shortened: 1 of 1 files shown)


playbooks/rfcs.md
```diff
@@ -39,6 +39,24 @@ Find the right repo:
...
+
+Decide who needs to know about this proposal, what role teammates will play in the process and how the RFC will be
+considered approved.
+
+- What is the scope of the proposal? Does the whole engineering team need to know or just a smaller group of
+  engineers?
+- Who should actively approve of this RFC? Who should give feedback?
+- Does the change make sense without further discussion from engineering leadership?
+
+Here are some examples of how an RFC could be resolved:
```
ashfurrow, line 54: I like the idea of adding an "implementation" to our RFC playbook, it makes things more tangible. I think the term "resolved" is a bit overloaded, though. To my mind, an RFC moves through the following sequence:

- Proposed (open for discussion)
- Resolved (based on discussion)
- Implemented

As it is, this PR treats "resolved" as a kind of "completed" state, but "resolved" just means that we've decided to do it (even if we haven't done it _yet_). Thinking about how we measure the success of an RFC upfront is a really good idea, but I think we should re-word this a bit to make it more obvious what "resolved" means.

williardx: Thanks for the feedback, that's a good point that "resolved" is not very clear. My intent here is to address RFC acceptance / approval and be able to clearly answer the question "How do we know we're done discussing?" for a given RFC.

admbtlr: Maybe it's better to stick with "approved", for the reasons that Ash mentions? It implies that everyone is in agreement, but that the thing in question has probably not yet been done, and as such makes a good halfway stage between "proposal" and "implementation".

iskounen:
**suggestion:** I think this RFC would set a good example for being more specific by making the following changes:

1.  Add a "How will this RFC be implemented?" section to the proposal, and explain who will be tasked with updating which RFC docs and the template.
1. Describe how we will establish a majority acceptance. How will we vote? When will we vote? Who will vote? etc.

playbooks/rfcs.md
joeyAghion: I like "resolved" because it covers both outcomes: in favor or against. But maybe we should just use the built-in language of "closed" for this?

playbooks/rfcs.md
```diff
@@ -39,6 +39,24 @@ Find the right repo:
 - Does it affect the whole dev team? (use [README](https://github.com/artsy/README))
 - Can't figure it out or needs to be private? (use [Potential][] as a fallback)
 
+Create a plan for how the RFC will be implemented. If the change is complex or large in scope consider defining a
```
joeyAghion, line 42: I'd use "resolved" or "closed" here instead of "implemented" (as commented below), since the latter implies the RFC is adopted.

playbooks/rfcs.md
```diff
@@ -65,14 +83,25 @@ Create an issue, and work from this template:
...
 
-Resolve the RFC, you can work with this template:
+To resolve the RFC you can work with this template:
 
     ## Resolution
-    We decided to do it.
+    We decided to do it. 50% of the engineering team actively approved of this change.
 
     ## Level of Support
     3: Majority acceptance, with conflicting feedback.
```
joeyAghion, line 107: (Commenting here only because I can't comment on the relevant lines below.)

This PR supersedes the recommendation below, so that can be removed:

> Resolving an RFC requires you to have some nuance about the feedback. If it seems to be unresolved, or still active a week later then calling a town hall style meeting for people involved will probably shake out some come kind of resolution.

playbooks/rfcs.md
```diff
@@ -65,14 +83,25 @@ Create an issue, and work from this template:
 
     You can see our discussion [in slack here](/link/to/slack.com)
 
-Give a week.
+    ## How this RFC is implemented
```
joeyAghion, line 86: Not especially related to this PR, but maybe this doc should just link to [the issue template](https://github.com/artsy/README/blob/master/.github/ISSUE_TEMPLATE/rfc.md) rather than replicating it. That template should also reflect your updates.

joeyAghion (approved):
I'm broadly in favor of clarifying and simplifying how RFCs are resolved. In that interest, can this be made even simpler? Maybe something like:

* Specify _whose_ feedback is necessary with @-mentions (or review requests?)
* Specify for _how long_ feedback will be collected and incorporated into the proposal. At least a week for feedback is appropriate for any proposal impacting all of engineering.
* Specify how approval will ultimately be decided. Does this depend on X% of the engineering team actively approving, point people of certain repos, or perhaps engineering leadership?

playbooks/rfcs.md
```diff
@@ -39,6 +39,24 @@ Find the right repo:
...
+- What is the scope of the proposal? Does the whole engineering team need to know or just a smaller group of
+  engineers?
+- Who should actively approve of this RFC? Who should give feedback?
+- Does the change make sense without further discussion from engineering leadership?
+
+Here are some examples of how an RFC could be resolved:
+
+- X% of the engineering team actively approves this change.
+- The point people of the impacted repos approve this change.
+- Engineering leadership approves the changes after collecting feedback from the team.
```
admbtlr, line 58: I feel like this list is pointing towards the larger and thornier problem of how large-scale change happens within the engineering org. If I'm writing a major RFC, how exactly do I choose which of your example resolutions is the right one? If I choose the first one, what should the value of X be? Who actually are the point people for the repos? What are the circumstances in which leadership needs to be involved? Who even is "leadership", in this context?

williardx: This issue is at the heart of this RFC. Right now there's no mechanism or guidelines for choosing any of this and it's up to the RFC creator to find the right way to build support for their RFC. In practice that means having a nuanced understanding of our org and knowing how support gets built and what the "right" kind of approval is for a given situation. It's largely tribal knowledge. So these suggestions are necessarily fuzzy because right now creating large change is ambiguous. I'm not able to answer your questions, and I don't necessarily even think that we should use these suggestions precisely because they're unclear. This RFC just takes a small step towards reducing ambiguity, though, by suggesting that we explicitly call out what the approval process is and debate it if necessary. But I'd love to see better-defined approval mechanisms or stronger guidelines for what to do when writing a major RFC!

playbooks/rfcs.md
williardx: Thinking on this more, we could update the template to be something like this:

```
How this RFC will be resolved

This RFC will be *accepted* if...
This RFC will be *rejected* if...
```

That gives some space to explicitly address the two outcomes (acceptance / rejection) and also think through how rejection might occur, which helps prevent ambiguous outcomes that don't feel like either. E.g., "This RFC will be considered rejected if it's not accepted within X days and there's no further discussion."

playbooks/rfcs.md
williardx: My intent with this section was to address more complex, potentially multi-stage RFCs. E.g., if we're doing an RFC that has a trial run of something before we roll it out across more systems / the org, the implementation section could detail how the trial run will work and how the RFC might change based on the results. But I can also see that being part of the "approval" process as well.

sweir27 (commented):
I'm also generally in favor of this but I think the proposed simplifications will increase the clarity of these changes!

playbooks/rfcs.md
williardx: How do you envision making a larger / more complex change happen within the RFC system? E.g. maybe a series of smaller, more targeted RFCs that build off of each other as we work towards the intended outcome? I could imagine, for example, an RFC to align on the outcome we want and to do a trial, and then another RFC to write up trial results / feedback and propose a full rollout.

williardx:
Thanks everyone for the feedback so far! I updated the PR with @joeyAghion's suggestions since I thought they were clear. I also removed the explicit implementation section since it was confusing, though I kept the suggestion to create a plan for how it'll be implemented. I'll merge this at the end of the week if there's no other feedback or suggested changes.

playbooks/rfcs.md
```diff
@@ -39,6 +39,19 @@ Find the right repo:
...
+Create a plan for how the RFC will be implemented. If the change is complex or large in scope consider defining a
+feedback process to iterate on the proposal as you implement and creating milestones and/or metrics to track
+progress.
+
+Decide how you'll resolve the RFC:
+
+- Specify whose feedback is necessary on this RFC with review requests.
+- Specify for how long feedback will be collected and incorporated into the proposal. At least a week for feedback
+  is appropriate for any proposal impacting all of engineering.
+- Specify how approval will ultimately be decided. Does this depend on X% of the engineering team actively
```
SamRozen, line 51: Not sure what the exact wording should be here but the goal of an RFC is not necessarily to build consensus. We need to make sure that the barrier to creating and pushing RFC is low, and in most cases an RFC with no feedback and no dissent that is "safe to try" should be good to go. 

So maybe the default approval criteria is: "Lack of dissent / No objection" ?

SamRozen (approved):
I'm supportive of this since it is helpful for complex / controversial changes. I also want to be mindful to keep the RFC bar low for simple / safe to try changes. Ultimately we'll make better decisions if we want more / smaller decisions more often. So I suggest that our default approval criteria for RFC is a lack of dissent / no objection.

playbooks/rfcs.md
williardx: That's a great suggestion, to have a default, low barrier approval model as part of this. I'll update the PR.

PR #350, opened 2020-12-10
fix: update documentation for resolving an incident

iskounen:
This PR updates our documentation for responding to incidents so that engineers are instructed to write postmortems for all incidents, regardless of severity.


dblandin (approved):
Thanks!

PR #352, opened 2020-12-15
chore: update practice names

ashfurrow:
It came up in standup on Monday that the practices have new names – let's update the docs 🙌 I also added back in the Front-End / Web practice.


ashfurrow:
Just updated the Notion doc, too 👍

SamRozen:
👍  Thanks Ash

PR #358, opened 2020-12-21
Mention the artsy cli command

jonallured:
This update to the docs introduces the artsy cli command for listing engineers for the on-call rotations. It's a little sparse on getting things all setup but that sorta seems outside the scope of this doc ya know? Anyway, this is better than nothing and we can continue to improve.


dblandin (approved):
Awesome 💯

PR #366, opened 2021-02-05
Suggest rule of thumb for what to do in practices vs. product teams

joeyAghion:
Tiny start to a "playbook" for practices.

As discussed, some other things this could document include:
* example agenda and list of candidate topics
* steps to create a practice Jira "board" via labels and saved filters

### DIFF
--- a/practices/README.md
+++ b/practices/README.md
@@ -3,6 +3,12 @@
 Practices are how we handle cross-cutting concerns across product teams, they are light-weight team structures
 which come together regularly to discuss.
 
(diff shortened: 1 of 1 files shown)


ansor4:
ah thanks for starting this!

SamRozen:
I wonder if we can be even more explicit here, for instance we're converging on a few areas where we expect the practices to be more involved in:
- Architecture / technical plan reviews
- Important technology choices
- Quality
- Performance
Trying to think of what else we usually expect the practices to cover.

PR #367, opened 2021-02-06
feat: run standup reminder via @artsy/cli every Monday at 9am ET

dblandin:
Use the Artsy CLI and scheduled GitHub actions for our Monday Informationals.

Currently in-progress. This doesn't support all of the Peril-backed tasks yet.

- [ ] Closed Source Rationale Check
- [ ] Compare Schemas
- [x] Standup Reminder
- [x] Weekly RFC Summary

#FF #FutureFriday

**Example**

Screen Shot 2021-02-07 at 3 33 17 PM

### DIFF
--- a/.github/workflows/mondayInformationals.yml
+++ b/.github/workflows/mondayInformationals.yml
@@ -16,6 +16,7 @@ jobs:
           echo "::set-output name=standup::$(artsy scheduled:standup-reminder)"
           echo "::set-output name=rfcs::$(artsy scheduled:rfcs)"
         env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           OPSGENIE_API_KEY: ${{ secrets.OPSGENIE_API_KEY }}
           SLACK_WEB_API_TOKEN: ${{ secrets.SLACK_WEB_API_TOKEN }}
       - name: Standup Reminder > Slack
--- a/.github/workflows/mondayInformationals.yml
+++ b/.github/workflows/mondayInformationals.yml
@@ -1,7 +1,7 @@
 name: "Monday Informationals"
 on:
   schedule:
-    - cron: "0 14 * * MON"
+    - cron: "15 1 * * MON"
 
 jobs:
   monday_informationals:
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Found a new file with underscores, this repo uses kebab-case for files: .github/workflows/monday_informationals.yml
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 501d40f3a071124ae30ef735440511034e8faa55
</p>

.github/workflows/mondayInformationals.yml
```diff
@@ -0,0 +1,33 @@
...
+  schedule:
+    - cron: "0 14 * * MON"
+
+jobs:
+  monday_informationals:
+    name: Monday Informationals
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Dependencies
+        run: yarn global add --silent @artsy/cli@alpha
```
damassi, line 12: Don't think it makes a big difference, but could we [use `npx` instead](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b) (which ships with node) and skip the install step?

dblandin: Ahh, that's a good idea. What if we were invoking the CLI multiple times within the same workflow? Would using npx install it N times?

damassi: Yup, it would, so this would only be useful for one-off commands

dblandin:
Going to merge this to test against the `@artsy/cli@alpha` version.

.github/workflows/mondayInformationals.yml
jonallured: Would be cool to generate a docker image with the cli already installed so this is moot ya know?

PR #368, opened 2021-02-08
RFC: Propose conventions for project set-up and configuration

joeyAghion:
This documents a convention for set-up and configuration that we've been trying to refine over several iterations in [Horizon](https://github.com/artsy/horizon/pull/177), Gravity ([1](https://github.com/artsy/gravity/pull/13345), [2](https://github.com/artsy/gravity/pull/13372)), [Volt](https://github.com/artsy/volt/pull/4582), and [Metaphysics](https://github.com/artsy/metaphysics/pull/2765).

The _aspiration_ is that setting up a new development environment is turn-key, and there's a shared, up-to-date place to put any configuration necessary for development.

### For debate:

**Whether to automate setup at all.** There's been a lot of discussion recently about whether project setup should be standardized and automatic or tailored and manual. Long term, containerized environments should solve this, but I tried to find a balance by suggesting:
* we establish default set-up scripts for the most common environment
* the setup be in-line documented, to serve as a guide in less common circumstances

**What "most common" environment to target.** Version manager and shell preferences vary widely, but we need to start somewhere. I took as defaults the tools currently supported by the shared setup script and working examples in projects. Ultimately I'm not attached to any of these, as long as there's always an "easy," default path.

### DIFF
--- a/playbooks/development-environments.md
+++ b/playbooks/development-environments.md
@@ -1,0 +1,96 @@
+---
+title: Development environments
+description: Getting set up to work on new projects
+---
+
+# Development environments
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Goals](#goals)
+- [Prerequisites](#prerequisites)
+- [Project set-up](#project-set-up)
+- [Shared configuration](#shared-configuration)
+- [Examples](#examples)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Goals
+
+- New working environments are set up quickly, easily, and consistently (ideally with a single command).
+- Team- or project-wide configuration changes are rolled out without needing a lot of communication and
+  coordination.
+- Projects adopting favored technologies leverage common, simple set-up.
+- Developers using common tools (mac, bash, homebrew, nvm...) get productive without any special set-up.
+- Developers choosing less common tools are still productive, if necessary using the basic set-up as a guide.
+- Projects default to configuration values that are appropriate for development (just as they run in "development"
+  mode by default).
+- Both local and containerized development are supported, with minimal duplication.
+- Sensitive values are never committed to repositories, whether open- or closed-source.
+- Other things being equal, less configuration, simpler-setup, and fewer dependencies are better than more.
+
+## Prerequisites
+
+Setup tooling is optimized for a default set of tools:
+
+- Mac computers
+- `bash` shell
+- [Homebrew](https://brew.sh/) for installing and managing supporting services
+- Docker for containerization
+- [nvm](https://github.com/nvm-sh/nvm) for managing node versions
+- [rbenv](https://github.com/rbenv/rbenv) for managing ruby versions
+- [hokusai](hokusai.md) for managing deployed applications
+- [VSCode](https://code.visualstudio.com/) for text editing
+- [awscli](https://aws.amazon.com/cli/) for AWS operations
+
+Developers who find other tools compelling may need to adapt setup procedures to their choices. For the sake of
+simplicity, we avoid making setup tooling heavily conditional on local choices.
+
+The [shared setup script](https://github.com/artsy/potential/blob/master/scripts/setup)🔒 installs these tools, and
+individual projects' setups may depend on them.
+
+These choices can and should be improved (by [RFC-style](rfcs.md) pull request), but care should be taken to update
+any dependent projects.
+
+## Project set-up
+
+Project's READMEs should clearly document the commands necessary for fresh set-up, ideally just reviewing and
+running a `./bin/setup` script (or whatever is idiomatic for that stack). The script should:
+
+- install dependencies such as data stores
+- install necessary language runtimes/versions
+- initialize configuration with values that are appropriate for development (ideally none)
+- install packages (rubygems, node modules, etc.)
+- create and seed databases, if appropriate
+- ideally be idempotent (i.e., re-running setup should work to update an existing environment)
+- be clearly organized and commented, so it can serve as a guide in incompatible environments
+
+## Shared configuration
+
+We strive to avoid committing any sensitive values to repositories, but sometimes they are necessary for local
+development. By convention, projects load configuration values from both:
+
+- a `.env.shared` file with common configuration values, and
+- a `.env` file with any developer-specific overrides (or empty)
+
+These files are excluded from source-control. The setup command initializes them, updating `.env.shared` from a
+private S3 location.
+
+Local ruby development depends on the [`foreman`](https://github.com/ddollar/foreman) utility and a `.foreman` file
+([e.g.](https://github.com/artsy/horizon/blob/master/.foreman)) to respect these files. Local node.js development
+depends on [node-foreman](https://github.com/strongloop/node-foreman) and the `--env` argument
+([e.g.](https://github.com/artsy/metaphysics/blob/edad4a5f2215a61bb09719901a4fdfd38cd0afcd/package.json#L19)).
+
+Docker-based development respects these files by listing both in docker's `env_file` property
+([e.g.](https://github.com/artsy/horizon/blob/2202391c9622b5ec655bf2c6d0f35ef379d0687f/hokusai/development.yml#L18-L20)).
+
+To _update_ shared configuration values, simply modify `.env.shared` and re-upload it to its shared location as
+`.env.<project>`. E.g.:
+
+    aws s3 cp .env.shared s3://artsy-citadel/dev/.env.zulu
+
+## Examples
+
+- [Gravity's `script/setup`](https://github.com/artsy/gravity/blob/master/script/setup)🔒
+- [Metaphysics' `scripts/setup.sh`](https://github.com/artsy/metaphysics/blob/master/scripts/setup.sh)
--- a/playbooks/README.md
+++ b/playbooks/README.md
@@ -25,6 +25,7 @@ practice should be incorporated, submit a PR!
 | [Data Migrations](/playbooks/data-migrations.md#readme) | Preparing, reviewing and executing data migrations |
 | [Dependency Management Recommendations](/playbooks/dependencies.md#readme) | How do we want to handle new dependencies to our apps? |
 | [Deployments](/playbooks/deployments.md#readme) | How systems are deployed at Artsy |
+| [Development environments](/playbooks/development-environments.md#readme) | Getting set up to work on new projects |
 | [Documentation](/playbooks/documentation.md#readme) | How and where to document or share content |
 | [Engineer workflow](/playbooks/engineer-workflow.md#readme) | How we work together |
 | [Extracting Services](/playbooks/extracting-services.md#readme) | When and why to consider extracting or decoupling systems |


playbooks/development-environments.md
```diff
@@ -0,0 +1,96 @@
...
+
+- a `.env.shared` file with common configuration values, and
+- a `.env` file with any developer-specific overrides (or empty)
+
+These files are excluded from source-control. The setup command initializes them, updating `.env.shared` from a
+private S3 location.
+
+Local ruby development depends on the [`foreman`](https://github.com/ddollar/foreman) utility and a `.foreman` file
+([e.g.](https://github.com/artsy/horizon/blob/master/.foreman)) to respect these files. Local node.js development
+depends on [node-foreman](https://github.com/strongloop/node-foreman) and the `--env` argument
```
mdole, line 82: @joeyAghion this bit surprised me - I've never even heard of `node-foreman`. Is that because it's so baked-in I've never noticed it/because I've mostly worked on Artsy apps that are Ruby on Rails-based?

jonallured:
I have opinions! But I do agree that there is value in writing this stuff down and being consistent. The things I want to advocate for:

* separate the idea of setting up a machine for Artsy development work from setting up an individual project
* project setup scripts should check for and fail when dependencies like postgres are not installed, they should not attempt to install postgres for me

FWIW, I have played around with using Docker for serious dev work and not had good experiences. Can someone demonstrate how this works well?

playbooks/development-environments.md
joeyAghion: It [is baked into](https://github.com/artsy/metaphysics/blob/245fd4e3740759b7077c438f6ba3f4ae5b771144/package.json#L19) commands like Metaphysics' `yarn dev`.

`dotenv` was the other utility we evaluated for this purpose, but we found that it doesn't support loading from multiple `.env*` files with correct precedence rules. (Having a shared source of common config and an isolated place for any individual developers' overrides was a goal of our work.)

joeyAghion:
>separate the idea of setting up a machine for Artsy development work from setting up an individual project

Totally agree. That's what I was aiming for by separating prerequisites (`aws`, `hokusai`, `brew`) in the base script from project-specific setup. Do you have something different in mind?

>they should not attempt to install postgres for me

I believe in your right to self-manage postgres, but I submit that there are _many_ cases where that trouble is unwelcome. A less experienced engineer might not already have a preferred method for setting it up. An onboardee or rotator might just be trying to get a seemingly small ticket done. (Similarly for someone visiting a project only briefly from another area of the platform.) Plus there's a significant benefit of many engineers setting up postgres the same way, which is that most set-up problems have already been encountered and solved.

That said, my ideal "middle ground" would be for set-up to prompt and allow developers to opt out of any specific step. I just couldn't find a way to do this that didn't depend on [verbose bash stuff](https://stackoverflow.com/questions/226703/how-do-i-prompt-for-yes-no-cancel-input-in-a-linux-shell-script) or a [whole ruby runtime](https://medium.com/braintree-product-technology/https-medium-com-braintree-product-technology-runbook-be6f072cfc0d).

> I have played around with using Docker for serious dev work and not had good experiences

Same here. There are improvements happening in the community (and over time, as everyone upgrades to 64GB machines 😛), but I didn't want us to hold our collective breath. This approach tried to be _compatible_ with docker-based development but treat local development in a first-class way since it will realistically be the dominant approach in the medium term.

jonallured:
We might be aligned then! What I'm saying is that the project-specific scripts should look like this:

https://github.com/artsy/pulse/blob/master/bin/setup

That script communicates that you need to have postgres setup, suggests the likely command you should run and fails to proceed. If we want to have a general laptop setup script that walks new hires through setting up things like postgres BEFORE they even run this Pulse script, then go for it. I'm just saying that project-specific setup scripts should not install postgres, just check that it is in fact installed. Agree?

joeyAghion:
Maybe postgres isn't the best example. Take elasticsearch, which multiple projects rely on but whose setup is more specialized than a simple command that can be printed to the console (in our case). Rather than print instructions for multiple finicky commands, I think the script can and should perform the installation.

jonallured:
> I think the script can and should perform the installation

I confess I'm getting a little lost here but I think what you're saying is that the general laptop setup script should setup elasticsearch, right? Or are you saying that the project script will do this?

joeyAghion:
Sorry I'm taking elasticsearch an example of a project-specific dependency (Gravity's, in this case) that has a somewhat involved setup. Personally I'd like the script to _execute_ the multiple steps required, rather than print out instructions or expect every engineer to be an expert at it.

jonallured:
Ok what I'm saying is that elasticsearch setup should be covered in the general laptop setup and the Gravity script should just verify it can see that the dependency is installed and if not direct one to the general laptop setup script.

joeyAghion:
Oh! I disagree... I think only developers who need elasticsearch should be running it. The alternative, of a general laptop setup installing the _union_ of all projects' dependencies, seems impractical.

dblandin:
**Managing System Dependencies**

> install dependencies such as data stores

I also lean towards "check but don't install system dependencies" (including data stores / languages).

This [Thoughtbot post](https://thoughtbot.com/blog/shell-script-suggestions-for-speedy-setups) example of a setup script structure looks nice, particularly the way in which it organizes those different steps of the setup.

The `bin/setup_steps/check_dependencies` script also has a "check and notify/fail" approach vs "check and install".

```bash
#!/bin/sh

./bin/setup_steps/check_dependencies
./bin/setup_steps/install_plugins
./bin/setup_steps/setup_ssl
./bin/setup_steps/pull_environment_vars_from_heroku

printf "${CLEAR_LINE}[6/6]🎉${GREEN}   Finished!${NO_COLOR}\n"
```

source: https://thoughtbot.com/blog/shell-script-suggestions-for-speedy-setups

---

> Sorry I'm taking elasticsearch an example of a project-specific dependency (Gravity's, in this case) that has a somewhat involved setup. Personally I'd like the script to execute the multiple steps required, rather than print out instructions or expect every engineer to be an expert at it.

For more complicated installations, one compromise might be to define a `bin/setup_steps/install_<dep>` script and prompt the user to run that if the dependency is not detected. This would give developers the option to quickly opt-in to an automated setup.

```
$ bin/setup
You must install and configure elasticsearch on your system before setup can continue
ℹ️   On macOS, you should run `./bin/setup_steps/install_elasticsearch`
```

dblandin:
@joeyAghion Should this PR title be prefixed with "RFC:"? I'm not sure if you want the automation (mainly Slack posts) which that would trigger.

joeyAghion:
I added `RFC` to the title. (Didn't realize that worked for PRs.)

To be clear, though, this is more than a proposal. This is an attempt to document an approach already piloted as part of the platform team's goals ~2 quarters ago, when "trouble setting up projects" was one of the frequently cited complaints about velocity. I've actually received good feedback about Volt and Metaphysics set-up "just working" since then, but--admittedly--it's not optimized for those with more curated environments.

I'll take a small detour into implementing escape hatches in our current reference examples, since this seems to be the main sticking point. However I don't want to expand the scope of this PR into deciding the ultimate setup script. I'd rather refine our shared examples over time as long as we agree about the goals and storage location for shared values.

(For what it's worth, [the Thoughtbot pattern](https://thoughtbot.com/blog/bin-setup) was a big influence on the current approach. The variant that breaks setup into multiple scripts seems complex to me at the moment.)

dblandin:
> I added RFC to the title. (Didn't realize that worked for PRs.)

Ahh, I think you're right. There might be some missing pieces there. Doesn't look hard to support though.

> I'd rather refine our shared examples over time as long as we agree about the goals and storage location for shared values.

Thanks for the context and definitely don't mean to scope creep. Documenting the current approach and refining over time sounds great to me 👍

playbooks/development-environments.md
```diff
@@ -0,0 +1,96 @@
...
+  mode by default).
+- Both local and containerized development are supported, with minimal duplication.
+- Sensitive values are never committed to repositories, whether open- or closed-source.
+- Other things being equal, less configuration, simpler-setup, and fewer dependencies are better than more.
+
+## Prerequisites
+
+Setup tooling is optimized for a default set of tools:
+
+- Mac computers
```
sweir27, line 37: We don't _need_ to call this out, but I'm curious if we have a current recommendation for computer specs for devs (i.e. 16 vs. 32GB of RAM)?

joeyAghion:
Short example of wrapping a setup step in a confirmation prompt (and could be improved with case insensitivity and other robustness):

```bash
read -n 1 -p "Install dependencies (elasticsearch, redis...)? [Y/n]: " answer
answer=${answer:-Y}
if [ "$answer" == "Y" ]; then
  brew bundle
else
  echo 'Skipping dependencies...'
fi
```

I don't think it's worth increasing the `bash` code required by 4x for this convenience. I'd rather the README say "review the `setup` script and choose to run it or perform the equivalent steps for your environment" and have the script look more like:

```bash
echo "Installing dependencies (elasticsearch, redis...)"
brew bundle
```

dblandin:
Here's another example without the interactive prompt:

```bash
deps=(redis-server mongod elasticsearch memcached postgres convert)
for dep in "${deps[@]}"
do
  if ! command -v "$dep" > /dev/null
  then
    printf "\`%s\` not found.\n\nRun \`brew bundle\` to install system dependencies.\n" "$dep"
    exit
  fi
done
```

And how that would look upon invocation:

```
$ scripts/setup
`redis-server` not found.

Run `brew bundle` to install system dependencies.
```

For anything more complicated that that, I feel like a separate `bin/install-dependencies` script would be worth the effort. If the initial setup phase doesn't pass all of the system dependency "checks", prompt to run the install script:

```
$ scripts/setup
Your Ruby version is 2.7.2p137, but .ruby-version specified 2.5.7.

Run `scripts/install-dependencies` to install system dependencies.
```

That install script can still be optimized for towards a particular set of assumptions: macOS, rvm, etc.

Generally, I'd advocate for the following phases:
- Check for system dependencies
- Install project dependencies (bundle/yarn)
- Run project setup (commit hooks / db setup+seed / env bootstrap)

> For what it's worth, [the Thoughtbot pattern](https://thoughtbot.com/blog/bin-setup) was a big influence on the current approach. 

From that post, it looks like the setup script _starts_ at the "Install project dependencies" phase and based on that other post, I don't think they bake system dependency installation into their setup scripts either. The examples I see _check_ and prompt for installation.

playbooks/development-environments.md
icirellik: I'm also quite surprised we are using `node-foreman`. Personally, I would drop it in favor of `dotenv` to reduce the complexity added by `node-foreman` and reduce the risk that someone attempts to use it in production.

joeyAghion: I welcome help with this, actually. I wasn't able to get `dotenv` to work the same way that docker's `env_file` does with the 2 files. That's the basic need.

joeyAghion:
I tried to get a version of this working and had the following experience:
* The commands to verify that a dependency is installed or a service available share quite a bit (constants, paths, version numbers...) with the code to actually perform that installation. As a result, splitting the verification and installation into separate files demanded _more_ and frequently _duplicate_ code.
* It's also just more complex to have setup split among multiple files, especially anticipating that many projects could follow these examples. For those reasons, I tried to introduce the short-circuiting you suggested while offering a choice to install within the same script.
* The actual commands to verify a dependency is available were frequently more complex than your example. E.g., homebrew might be running a service but the associated executables aren't necessarily available in the `PATH`. This is solvable; just the end result isn't as concise.
* Since the commands to check a dependency are pretty different, I saw no need to lump all dependency installations together (e.g., re-installing postgres if the expected ruby version isn't available).

[The result](https://github.com/artsy/gravity/compare/joeyAghion:setup) works, but requires a lot more bash scripting so I'm not in love. To my simple mind, the more bash scripting is involved, the more fragile and buggy the script might be, the less useful it will be, and ultimately the less maintained and current it will be. I'd rather keep the script as concise as possible but caveat it in READMEs like:

> Review the [setup script](link) to understand this project's needs, then run it or perform the equivalent steps for your environment.

@jonallured @dblandin: you've been most vocal but also [I suspect] tailor your environments heavily. This convention doesn't aspire to replace your local choices. 

Others with more "off the rack" development environments: any opinions about which setup examples achieve the right balance of robustness and maintainability?

joeyAghion:
### Resolution

There was acceptance of the _goals_ and _patterns_ introduced in this document, but conflicting opinions about the implementation convention.

### Level of Support

3: Majority acceptance, with conflicting feedback.

### Additional Context:

The examples should _definitely_ be taken as a work in progress (and we should update this document as they're refined). The document has also been adapted slightly to recommend reviewing and running equivalent set-up commands depending on each developer's preferences and environment. This is a stop-gap until docker or something else unifies default development environments.

### Next Steps

In addition to the projects that have piloted this approach, work is in progress in Pulse (https://github.com/artsy/pulse/pull/877) and planned for Force, which have both been plagued by set-up complaints.

PR #378, opened 2021-03-19
fix: Update Technical Leads

SamRozen:
Updating Technical Leads section in our career ladder to more accurately reflect the current responsibilities.


jonallured (approved):
That podcast episode I did with @joanna20 might be good to drop here but we don't really have episode pages yet. 😕  /cc @pepopowitz

brainbicycle (approved):

PR #402, opened 2021-08-16
Add a prompt for new podcast episodes

anandaroop:
Noticed while running today's standup that we could benefit from a prompt for new episodes, esp. since we've been so prolific lately.


mdole (commented):
Yay thanks! #mergeongreen

PR #404, opened 2021-08-26
docs: Minor grammar and clarity improvements

jacobherrington:
Some of these tweaks are based solely on my opinions, but I felt like I should suggest these while reading these documents!


MounirDhahri (approved):
Thank you so much for making these changes 🙌

PR #406, opened 2021-09-03
docs: Adding the postcards channel to the slack list of other topics

chr-tatu:
Small PR to add the postcards channel to the slack list of other topics channels


shaq31415926:
WOOH! GREATEST SLACK CHANNEL EVER

PR #409, opened 2021-09-13
doc: correct midnight on on-call SLO diagram

starsirius:
Update midnight on the SLO diagram from 12pm to 12am.


sweir27:
Beautiful! 💅

PR #411, opened 2021-09-21
feat: updated scheduling process

mc-jones:
This PR updates the on-call scheduling process. The primary change is that the schedule will now repeat rotations indefinitely. The goal is to remove the need for manual scheduling every few months or so. In support of this, we've added a section on how to add new engineers, which aims to limit the disruption to the short-term schedule.

### Considerations
- This process will require engineering managers to add new employees to the schedule, which will require a few checks (see the "Considerations when adding a new engineer" section).
- Similarly, remove an engineer will require the engineering manager to take the on-call spot of the departing engineer (if they haven't already completed their turn in the current rotation), but this is consistent with the existing process. 

### Next Steps
- Make the next schedule. The current schedule ends October 18.
- Monitor and smooth out any issues that pop up. 

This is part of a larger epic aiming to improve the on-call process:
- https://artsyproduct.atlassian.net/browse/PLATFORM-3495


playbooks/support/scheduling.md
```diff
@@ -1,74 +1,45 @@
 # Scheduling
 
-The [Engineering On-Call schedule][schedule] is configured in OpsGenie. Trading shifts (because of vacations,
-obligations, etc.) is encouraged as long as the schedule is kept up-to-date. Please address any scheduling issues
-as early as possible.
+The [Engineering On-Call schedule](https://artsy.app.opsgenie.com/teams/dashboard/ee381004-a72e-42ef-a733-b350d6693c6c/main) is configured in OpsGenie and works on a rotation basis. Each engineer will be on-call for one week per rotation. A rotation lasts for the period of time it takes for each participant to be on-call once, and will repeat itself indefinitely.
 
-A month before the next round of on-call is slated to begin (can check when the current round ends to verify this
-timeline), we being scheduling the next round.
+Trading shifts (because of vacations, obligations, etc.) are encouraged as long as the schedule is kept up-to-date. Please address any scheduling issues as early as possible.
```
sweir27, line 5:
```suggestion
Trading shifts (because of vacations, obligations, etc.) is encouraged as long as the schedule is kept up-to-date. Please address any scheduling issues as early as possible.
```

playbooks/support/scheduling.md
```diff
@@ -1,74 +1,45 @@
...
 
 ## Steps for Scheduling On-Call
 
-We use a staggered schedule to retain context and limit disruption as engineers shift in/out of rotation. "A"
-rotations start/end on a Monday at 11am. "B" rotations start/end on a Wednesday at 11am.
+We use a staggered schedule to retain context and limit disruption as engineers shift in/out of rotation. "A" rotations start/end on a Monday at 11 am. "B" rotations start/end on a Wednesday at 11 am. Each engineer will be on-call once per rotation.
 
-1. Generate a list of current engineers. This should include everyone who started before the next round is supposed
-   to begin (can use [Team Navigator][] as a reference). Bonus points for using [artsy-cli][]:
+1. Generate a list of current engineers. This should include everyone who started before the next round is supposed to begin (you can use [Team Navigator](https://team.artsy.net/) as a reference).  Bonus points for using [artsy-cli](https://github.com/artsy/artsy-cli):
```
sweir27, line 11:
```suggestion
1. Generate a list of current engineers. This should include everyone who started before the next round is supposed to begin (you can use [Bamboo](https://artsy.bamboohr.com/anytime/directory.php) as a reference).  Bonus points for using [artsy-cli](https://github.com/artsy/artsy-cli):
```

Since I believe we've deprecated Team Nav!

playbooks/support/scheduling.md
```diff
@@ -107,5 +78,28 @@ To override a shift:
...
+   <img width=500 src="images/opsgenie-addengs-1.png" />
+   <img width=500 src="images/opsgenie-addengs-2.png" />
+
+2. Add the engineer to the end of the participants list. Their position in the rotation will match their position in the participants list on the rotation edit page, for example:
+
+Considerations when adding a new engineer:
+
+- You should aim to avoid pairings with two new engineers and you can place the engineer earlier in the rotation to avoid this.
+- **NOTE: If you add engineers anywhere other than the end of the rotation, the schedule of the engineers who have not completed their shifts will be modified, as the order of the participants matters. You should therefore update the affected engineers to avoid scheduling conflicts.**
+- A new hire should be given 60 days after their start date to get the necessary context before having an on-call shift.
```
sweir27, line 96: Should managers add engineers to the rotation right when they join (but ensure that they aren't on the schedule for 60 days), or should they wait until they've been at Artsy for 60 days and then add them to the rotation?

I'm not sure if it matters, but I'm sure that people will ask. 😄

playbooks/support/scheduling.md
```diff
@@ -107,5 +78,28 @@ To override a shift:
...
+
+   <img width=500 src="images/opsgenie-addengs-1.png" />
+   <img width=500 src="images/opsgenie-addengs-2.png" />
+
+2. Add the engineer to the end of the participants list. Their position in the rotation will match their position in the participants list on the rotation edit page, for example:
+
+Considerations when adding a new engineer:
+
+- You should aim to avoid pairings with two new engineers and you can place the engineer earlier in the rotation to avoid this.
+- **NOTE: If you add engineers anywhere other than the end of the rotation, the schedule of the engineers who have not completed their shifts will be modified, as the order of the participants matters. You should therefore update the affected engineers to avoid scheduling conflicts.**
```
sweir27, line 95: Can we have some guidelines around this?

Like-- How far in advance should we assume the schedule is "set"? If the protocol is to inform engineers whose schedules may have changed, can we provide a sample slack message?

playbooks/support/scheduling.md
```diff
@@ -107,5 +78,28 @@ To override a shift:
...
+- A new hire should be given 60 days after their start date to get the necessary context before having an on-call shift.
+- Check for an unbalanced number of rotation participants in rot-a vs rot-b. If staff locations result in one rotation having significantly fewer engineers, you should prioritize a balanced rotation schedule over Eastern Time Zone coverage.
+
+### Removing  Engineers
+
+When an engineer leaves:
+
+- They will be replaced with a no-one placeholder in the schedule as part of the [off-boarding checklist](https://docs.google.com/document/d/10mmqkXnYVp0ZOmF5JwVPyxkzcyIN-rqO8prHgIi11lw/edit) and replaced with a no-one placeholder
```
sweir27, line 103: Does the "no-one" placeholder only last for one rotation, or does it stay in the schedule? I'm wondering if we also need a process around removing the "no-one"?

playbooks/support/scheduling.md
mc-jones: This is a good call out. My initial thinking was that they should do it right away. But if we are within the 60-days  or a rotation completing adding them to the end would violate this guidance. I will update to have managers delay adding until a new rotation has started if within the 60 days.

playbooks/support/scheduling.md
mc-jones: Since we won't add new engineers within 60 days of the rotation completion, we can assume the schedule will no longer shift (other than individual swaps) at this point.

And will add a sample slack message! Good call.

playbooks/support/scheduling.md
mc-jones: Another great call-out. We definitely do as it will add to all subsequent rotations too. Will add.

playbooks/support/scheduling.md
```diff
@@ -107,5 +79,49 @@ To override a shift:
...
+
+- You should aim to avoid pairings with two new engineers and can place the new engineer earlier in the rotation to avoid this.
+- **NOTE: If you add engineers anywhere other than the end of the rotation, the schedule of the engineers who have not completed their shifts will be modified, as the order of the participants matters. You should therefore update the affected engineers to avoid scheduling conflicts. You can message them something like:**
+
+   > Hi @engingeer-whose-schedule-shifted! We've added a new engineer to the on-call schedule which has modified the dates of your on-call shift. Please take some time to review the schedule and trade shifts as needed.
+
+   - In certain scenarios, to minimize shifting multiple engineers schedules, it may be preferable to reach out to one engineer directly and ask them to switch shifts. Something like:
+
+      > Hi @engineer! We've added a @new-engineer to the schedule and would like to support them by not pairing them with another new engineer for their first on-call shift. Would you be able to switch shifts with them? The new dates of your on-call shift would be X to Y.
+
```
joeyAghion, line 113: I'd like this whole section to be more fool-proof somehow.
* Can we advise adding new members immediately to (a) the first `no one` slot that's >60 days out, (b) the end of the rotation if >60 days out, or (c) in the case the rotation is ending sooner, repeating (a) and (b) once the new rotation starts?
* If we do that, can we altogether avoid shifting near-term rotation schedules, and the necessary communication?

playbooks/support/scheduling.md
```diff
@@ -107,5 +79,49 @@ To override a shift:
 
    <img width=500 src="images/opsgenie-override-2.png" />
 
-[schedule]: https://artsy.app.opsgenie.com/teams/dashboard/ee381004-a72e-42ef-a733-b350d6693c6c/main
-[artsy-cli]: https://github.com/artsy/artsy-cli
+## **Adding Engineers**
+
+Generally, new engineers should be added to the end of the schedule rotation associated with their time zone. To do this:
```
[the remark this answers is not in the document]
mc-jones: The thinking was that we would not utilize `no one` slots beyond what is laid out in the `Removing Engineers` section. 

>When an engineer leaves:
>- They will be replaced with a no-one placeholder in the schedule as part of the [off-boarding checklist](https://docs.google.com/document/d/10mmqkXnYVp0ZOmF5JwVPyxkzcyIN-rqO8prHgIi11lw/edit).
>- The user’s manager will be asked (in #dev-managers) to override that shift with themselves.
>- Their Opsgenie user account will be deleted.
>
>Responsibility of the departing engineer's manager:
>
>- Once an engineering manager overrides the departing engineer's next shift, they should remove the no-one placeholder from the participants list.

playbooks/support/scheduling.md
mc-jones: See above comment. Happy to switch to this if we feel this is a better approach than removing the no-one placeholders.

But let me know if my point is unclear.

playbooks/support/scheduling.md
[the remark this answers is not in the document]
joeyAghion: > Once an engineering manager overrides the departing engineer's next shift, they should remove the no-one placeholder from the participants list.

This^ must mean once the overridden shift finishes, right? Otherwise, I think removing the `no one` placeholder will shift the schedule, despite the manager's override.

playbooks/support/scheduling.md
joeyAghion: My thinking is:
* Even with this approach, there _will_ be `no one` placeholders in the schedule (between engineers departing and their next [overridden] shift). If they're there (until the manager returns after the shift to remove them), why not take advantage of them to randomize the new additions?
* Otherwise, won't we necessarily lump all of the newest team members together at the end of the rotation necessitating various judgment calls, communication, and adjustments?

joeyAghion (approved):
I tried to explain my concerns, but you can decide how best to proceed since you're most familiar with the particulars.

playbooks/support/scheduling.md
mc-jones: As mentioned above, I agree this makes more sense, give the shifting that happened when the no-one placeholder is removed. I'll incorporate the suggestions you make here.

playbooks/support/scheduling.md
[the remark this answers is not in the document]
mc-jones: Hmm. Somehow missed this. You're correct. It does still shift it down. With that, I agree it does make sense to utilize no-one placeholders, rather than remove them. I'll adjust to include the recommendation made below.

playbooks/support/scheduling.md
mc-jones: @joeyAghion I Have incorporated the changes but left the 'NOTE' bullet. I want to explicitly call out there is a cascading effect if the event an engineering manager does need to put in a new location for some reason.

PR #423, opened 2021-11-17
feat: notifications playbook

admbtlr:
Adds a playbook on how to set up notifications so that they don't disturb you outside of working hours.


kathytafel (approved):
This looks great, thanks for making a playbook. Super helpful and we can add other tips over time.

PR #425, opened 2021-11-19
Add steps for migrating deployed systems to use the main branch

joeyAghion:
Doppler's switch to use `main` (https://github.com/artsy/doppler/pull/218) was straightforward, but revealed some further steps that will be necessary when updating deployed systems.

(Please ignore the unrelated prettier-forced updates.)


kajatiger (approved):
❤️

PR #429, opened 2021-12-08
Update engineering's team headcount

fladson:
The number was taken from the employee's directory (filtering by Engineering).


fladson:
This is what I see when grouping by engineering.
Screenshot 2021-12-08 at 16 41 50

Please ignore/close and apologies if the number is not accurate.

PR #431, opened 2021-12-10
doc: add a note for external tools that can possibily get disrupted after …

sonecabr:
This PR do a small improvement on the playbook for branch renaming by adding external tools that possibily get disrupted after rename master branch to main.

### DIFF
--- a/playbooks/rename-master-to-main.md
+++ b/playbooks/rename-master-to-main.md
@@ -129,3 +129,9 @@ project's "stages":
 
 - Rename the `master` stage to `main`
 - Input `main` as the branch name as well
+
+### External tools
+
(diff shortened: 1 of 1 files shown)


kathytafel (approved):

playbooks/rename-master-to-main.md
```diff
@@ -129,3 +129,9 @@ project's "stages":
 
 - Rename the `master` stage to `main`
 - Input `main` as the branch name as well
+
+### External tools
+
+Since we haver external tools that are not possible to map by checking the repository code itself, this is something we need to consider as a Team duties after rename as well.
```
pvinis, line 135:
```suggestion
Since we have external tools that are not possible to map by checking the repository code itself, this is something we need to consider as a Team duties after rename as well.
```

PR #433, opened 2021-12-10
docs: Update rename master to main docs

gkartalis:
# Description

While we were in the FF Berlin Mob inclusive language zoom call we were discussing about updating the CI and docs before updating the actual branch since the open pr might take some time before getting approved, so I am moving that up on the steps of renaming master to main.


MounirDhahri (approved):
🎉

olerichter00 (approved):

PR #439, opened 2022-01-28
Propose Unleash trial (for feature-flagging and split-testing) more formally via radar

joeyAghion:
There's been a lot of interest in feature-flagging and split-testing recently. A few trials of specific tools are in intermediate states while we evaluate or unwind them, so let's make that more clear in our tech radar. Specifically:

* **Trial Unleash for feature-flagging and split-testing**. Support has been added to 2 back-end systems so far, and work is in progress (or planned) to integrate it into clients like Force and Eigen. Integrating into multiple clients is worthwhile despite not yet being fully "adopted," because how the service operates across the platform and stacks is an important element of this evaluation. There's some risk of wasted work, but little long-term maintenance burden because client SDKs fall back to defaults when the service is unavailable. <Debate...?>
* **Avoid new `LabFeature`s** since they're easily replaced by Unleash's built-in "user ID strategy."
* **Avoid new `SiteFeature`s** since they're easily replaced by Unleash's built-in "standard strategy."
* Document that **we strive to split-test product changes** in parallel, rather than simply monitoring metrics after the fact. Due to the seasonality and event-sensitive nature of our product, retroactive monitoring is hardly enough to confirm that product improvements are actually that.

The interactive version of these changes can be viewed [here](https://radar.thoughtworks.com/?sheetId=https%3A%2F%2Fraw.githubusercontent.com%2Fartsy%2Freadme%2FjoeyAghion%2fradar-unleash%2Fplaybooks%2Ftechnology_radar%2Fartsy-tech-radar.csv).

**Q:** Should we document Split.io's "hold" status here, or is that already obvious enough to anyone involved?

Note: other diffs are automated and unrelated.

### DIFF
--- a/onboarding/README.md
+++ b/onboarding/README.md
@@ -15,6 +15,7 @@ More onboarding resources:
 <!-- start_toc -->
 | Doc | Overview |
 |--|--|
+| [Engineering Team Introduction](/onboarding/engineering-introduction.md#readme) | A brief introduction |
 | [Onboarding Responsibilities for Managers of New Hires](/onboarding/managers.md#readme) | Things for managers to remember during the early days of onboarding |
 | [Onboarding Responsibilities for Mentors of New Hires](/onboarding/mentors.md#readme) | How to be an effective mentor of a new hire |
 | [Onboarding for New Hires](/onboarding/new-hires.md#readme) | Your first steps to being productive |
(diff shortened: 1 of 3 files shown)


kathytafel:
Document the hold.
Let's make sure we are also documenting the how to track part.

starsirius (approved):
Very excited! 🎉

dblandin (approved):
Sounds good to me! 👍🏼

PR #443, opened 2022-02-08
Document our recommendation for continuous deployment

joeyAghion:
We never formally updated this after last quarter's successful push to get almost all projects deploying continuously.

This PR documents the current preference for continuous deployment and provides some rationale.


rumyantseva (approved):
LGTM! Thanks for clarifying it!

ovasdi (approved):
LGTM

PR #446, opened 2022-02-25
docs: update incident-handling playbook - close alerts when incident is over

rumyantseva:
Added a line to close alerts as part of incident management.


dblandin (approved):
Looks great! Thanks so much for updating this! 🎉

PR #451, opened 2022-03-07
doc: add documentation around adding new config variables

fladson:
Adding some more documentation about environment variables.
For more context, see [thread](https://artsy.slack.com/archives/CA8SANW3W/p1646681796877969).

### DIFF
--- a/playbooks/development-environments.md
+++ b/playbooks/development-environments.md
@@ -90,6 +90,11 @@ To _update_ shared configuration values, simply modify `.env.shared` and re-uplo
 
     aws s3 cp .env.shared s3://artsy-citadel/dev/.env.zulu
 
+If you are _adding_ a new configuration value, you need to run additional cmds:
+
+    hokusai [staging|production] env set NEW_ENV_VARIABLE=<value>
+    hokusai [staging|production] refresh
+
 ## Examples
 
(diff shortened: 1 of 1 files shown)


mc-jones:
Thanks for this @fladson! So do want to draw a distinction between the development environment and the staging and production runtime environment. For development, the `env.shared` should be sufficient. But if you want to test those changes in staging or see them in production, then that is when you need to use `hokusai ... env set ...`. This is mentioned here: https://zoom.us/j/93825693217, so wondering if better to note as a follow up and link to the existing doc? What do. you think?

fladson:
> Thanks for this @fladson! So do want to draw a distinction between the development environment and the staging and production runtime environment. For development, the `env.shared` should be sufficient. But if you want to test those changes in staging or see them in production, then that is when you need to use `hokusai ... env set ...`. This is mentioned here: https://zoom.us/j/93825693217, so wondering if better to note as a follow up and link to the existing doc? What do. you think?

Thanks @mc-jones for the "review", now I realised that this refers only to the local environment, sorry.

I guess you meant this [doc](https://github.com/artsy/README/blob/3b82f9d0853e55ec7ebba76d1f50dc5e386ac18d/playbooks/hokusai.md#updating-environment-variables), right?

I updated the change to what you suggested.

mc-jones:
Sorry about the delayed merge on this!

PR #453, opened 2022-03-10
Chore: Adding google groups link to the New Engineers responsibilities section

Mitch-Henson:
Directly calling out the google groups URL for those unfamiliar with how to add themselves to a group


gkartalis (approved):
🥇 🚀

PR #454, opened 2022-03-14
chore: add peer learning groups prompt

admbtlr:


jo-rs (approved):
Didn't even know this file existed.
Thank you for adding, and sorry for being annoying 😅

pvinis:
this file is basically a dup of https://www.notion.so/artsy/Month-Day-Year-fbc9df949fef430a9f9359f34158dec6. should we point from one to the other at least?

jo-rs:
> this file is basically a dup of https://www.notion.so/artsy/Month-Day-Year-fbc9df949fef430a9f9359f34158dec6. should we point from one to the other at least?

Seeing as this repo is open source and the notion isn't, probably not? 🤔  
But, still, annoying that we end up with discrepancies.

admbtlr:
Oh that's interesting I've never even looked at that Notion (guess I never took standup notes). Maybe it's true that we should kill this and keep that. Although actually now I just realised that this file is linked to from 4 other places in the repo, so maybe we should just live with the drift for now.

PR #460, opened 2022-04-28
Update rfc.md

patrinoua:
Add some more info and links on the RFC template


mdole (approved):
Nice! Definitely helpful for future RFC writers

PR #464, opened 2022-05-04
Rotation Docs Updates

chr-tatu:
**Changes Summary in sprint-rotation.md**
- Adding Rotation buddy responsibilities 
- Removing forced line wraps.
- Adding a suggestion to organize 1:1 with different team members
- Reducing the heading by 1, since the document only contained Header 2+.

### DIFF
--- a/onboarding/sprint-rotation.md
+++ b/onboarding/sprint-rotation.md
@@ -5,45 +5,29 @@ description: What is the sprint rotation and how does it work?
 
 ## Sprint Rotation
 
-Once an engineer has gone through general onboarding and done some basic set-up, they'll spend the following few
-sprints rotating through a variety of engineering teams.
+Once an engineer has gone through general onboarding and done some basic set-up, they'll spend the following few sprints rotating through a variety of engineering teams.
 
 ### Why rotate?
 
-We want engineers to be familiar with the full breadth of products, systems, and people at Artsy, but at a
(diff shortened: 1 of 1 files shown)


joeyAghion (approved):
Last I checked, the line-wrapping was enforced by an automatic pre-commit hook. (However, I'm not a fan, so if you can get rid of that, or if I just have an old setup and can be ignored, that would be great.)

chr-tatu:
> Last I checked, the line-wrapping was enforced by an automatic pre-commit hook. (However, I'm not a fan, so if you can get rid of that, or if I just have an old setup and can be ignored, that would be great.)

It makes sense to me, too. If someone wants a line wrap, it can be defined when viewing the file, not embedded in a file. I checked `.git/hooks` and there is no commit hook defined for this repo.

lidimayra:
On the line wrapping topic, it's worth noting that a related discussion also happened here:  https://github.com/artsy/potential/issues/443

joeyAghion:
The git hook (at least in my case) is also responsible for generating tables of contents (and maybe other things). See [the precommit command](https://github.com/artsy/README/blob/33ab04f1bb3ffe9a609ec4065f580b39adf787c9/package.json#L16).

I think this could merit some additional work (not related to this PR!), maybe just a `setup` script that sets up any necessary hooks, excluding the line-breaking.

chr-tatu:
That's interesting. I didn't know about it. I'll add it to my todo list, @joeyAghion .

PR #467, opened 2022-05-31
rfc: use PRs instead of issues for RFCs

admbtlr:
Proposing a new approach for RFCs.

### DIFF
--- a/RFCs/use-prs-instead-of-issues.md
+++ b/RFCs/use-prs-instead-of-issues.md
@@ -1,0 +1,8 @@
+# Use PRs instead of issues for RFCs
+
+We already have a [playbook](../playbooks/rfcs.md) for RFCs. We used to use GitHub [issues](https://github.com/artsy/README/issues?q=is%3Aissue+is%3Aclosed), but this approach has two drawbacks: 
+
+- it's impossible to add inline or threaded comments to an issue, and the conversation can become very hard to follow
+- it's impossible to tell from the list of closed issues which RFCs were accepted, and which rejected
+
+Changing our process to use PRs that create documents within an RFC directory solves these two problems.
--- a/RFCs/use-prs-instead-of-issues.md
+++ b/RFCs/use-prs-instead-of-issues.md
@@ -18,3 +18,5 @@ We will use PRs for all RFCs from now on. I have updated the [RFC Playbook](/art
 
 #### Exceptions
 RFCs in individual repos should continue to use Issues, unless the team decides to add an rfc directory to a repo and start storing docs there.
+
+Discussion: https://github.com/artsy/README/pull/467
--- a/RFCs/use-prs-instead-of-issues.md
+++ b/RFCs/use-prs-instead-of-issues.md
@@ -6,3 +6,15 @@ We already have a [playbook](../playbooks/rfcs.md) for RFCs. We used to use GitH
 - it's impossible to tell from the list of closed issues which RFCs were accepted, and which rejected
 
 Changing our process to use PRs that create documents within an RFC directory solves these two problems.
+
+## Resolution
+We decided to do it. 25% of the engineering team actively approved of this change; no-one disapproved.
+
+## Level of Support
+1: Overwhelming positive feedback.
+
+## Next Steps
+We will use PRs for all RFCs from now on. I have updated the [RFC Playbook](/artsy/README/blob/main/playbooks/rfcs.md) accordingly.
+
+#### Exceptions
+RFCs in individual repos should continue to use Issues, unless the team decides to add an rfc directory to a repo and start storing docs there.
--- a/playbooks/rfcs.md
+++ b/playbooks/rfcs.md
@@ -20,12 +20,12 @@ process with as much bang for buck as possible given our size.
 
 ### TLDR: Our process looks like:
 
-- Someone opens an issue on a repo that begins with "RFC: ", and that causes a message to be posted to Slack's #dev
-  room via Peril.
+- Someone opens a pull request on a repo (or an issue, if it's a repo that doesn't have an rfc directory to hold RFC docs) that begins with "RFC: "; this causes a message to be posted to Slack's #dev
+  channel via Peril.
 - The RFC proposer announces the RFC in the next weekly dev standup.
 - At least a week goes by for discussion. After this the proposer uses the resolution template to summarize the
   outcome.
-- Changes are either enacted, or not based on the discussion.
+- Changes are either enacted or not, based on the discussion.
 
 A lack of response from others is assumed to be positive indifference.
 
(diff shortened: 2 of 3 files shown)


RFCs/use-prs-instead-of-issues.md
```diff
@@ -0,0 +1,8 @@
+# Use PRs instead of issues for RFCs
+
+We already have a [playbook](../playbooks/rfcs.md) for RFCs. We used to use GitHub [issues](https://github.com/artsy/README/issues?q=is%3Aissue+is%3Aclosed), but this approach has two drawbacks: 
+
+- it's impossible to add inline or threaded comments to an issue, and the conversation can become very hard to follow
```
pvinis, line 5: using this worked

> wow

but yea, that totally makes sense, cause now we can have many threads running!

admbtlr:
> One thing i would like to have defined/cleared up, is if we should use the PR desc or just keep everything on the PR files. I think PR files makes more sense, but I would like it if all of us follow one way.
> 
> Another thing, is about how we handle changes/reviews. In my mind, it makes sense to **only** leave comments on the PR and comments on the PR files/lines. It seems strange to use approvals/request changes. Seems like the author should be the one responsible to either merge or close, with the final comment of the next steps. Is that what you have in mind? How are you thinking of all that?

The outcome of an RFC should be the new markdown document that the PR is proposing to add. This document represents the content of the RFC and, once merged, is itself a permanent record of the org's decision to accept and act on the RFC. 

So yes, the discussion should take place within the context of the document, and the PR description can be left empty. Or just include a brief summary of the document, as I did here.

And yes, it's probably easier to use 👍 / 👎 reactions on the PR description, rather than approving / requesting changes. And yes, it should be the author who merges or closes the PR, based on the votes and discussion.

patrinoua:
love this 🔥

damassi:
It just makes an enormous amount of sense 👍 Nice ideal @admbtlr.

jacobherrington:
> The outcome of an RFC should be the new markdown document that the PR is proposing to add. This document represents the content of the RFC and, once merged, is itself a permanent record of the org's decision to accept and act on the RFC.

100% agree with this line of thinking.

joeyAghion:
I'm very in favor of RFCs as PRs (and we've [had a few](https://github.com/artsy/README/pulls?q=is%3Apr+rfc+is%3Aclosed+label%3ARFC)). My preference is for the outcome to not just be a new doc in the `RFCs/` directory, but corresponding updates to our playbooks, tech radar, etc. _That's_ where the meat of significant process changes can be debated, and where a developer will look for succinct documentation rather than the whole of our RFC history. (Of course they can all be linked together, too.)

gkartalis:
Love this idea, it makes so much sense! The best feature of this approach is the ability to let people provide in-context suggestions or changes! Thanks for bringing this up!

RFCs/use-prs-instead-of-issues.md
```diff
@@ -0,0 +1,8 @@
+# Use PRs instead of issues for RFCs
```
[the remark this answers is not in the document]
gkartalis: > They're a whole new medium that we're not currently using, so there would be some cognitive overhead involved in introducing them (Having said that, maybe we should be using them more widely? Are there other needs we have that could be met with discussion? I think there's too much overlap with our Slack usage, and that discussions are best for wider communities rather than internal company discussions, but could be worthwhile investigating)

100% agree with this we don't need to put congitive overhead to anyone 👍 but worth investigating maybe at some point in the future? 🤔 

> You can't do inline, line-level comments - which is the main motivation for changing our approach in the first place

Well, we might be able to do it! Not with the way that we can do it on PRs but rather with the quote reply format ^ (pasting the line/code that we want to comment on) and the good thing with discussions is that we can start a thread from the comment like [this one](https://github.com/vercel/next.js/discussions/16832#discussioncomment-61499) that could act as an inline comment. 

Don't consider [this comment](https://github.com/artsy/README/pull/467#discussion_r899107249) blocking for this RFC, just wanted to put this into our radars because it looks super interesting!

admbtlr:
> (@admbtlr seems like we should update the Markedown document this RFC introduces to include that part, right? 😉).

I added a little more detail in a new commit - does that cover it @mdole ?

mdole:
LGTM! Thanks @admbtlr

playbooks/rfcs.md
```diff
@@ -20,12 +20,12 @@ process with as much bang for buck as possible given our size.
 
 ### TLDR: Our process looks like:
 
-- Someone opens an issue on a repo that begins with "RFC: ", and that causes a message to be posted to Slack's #dev
-  room via Peril.
+- Someone opens a pull request on a repo (or an issue, if it's a repo that doesn't have an rfc directory to hold RFC docs) that begins with "RFC: "; this causes a message to be posted to Slack's #dev
```
erikdstock, line 23: I'm a little confused here so want to check my understanding:
- We can create RFCs from any repository, defaulting to an `/rfc` directory
- The rfc directory on the main branch is effectively a list of passed RFCs

Should engineers therefore view the rfcs directory as a directory of policies and decisions we've made, or should a given PR include changes elsewhere as well?

admbtlr: Your understanding is correct, but as far as I understand it, you're asking whether an RFC PR should also include any actions that may be taken to resolve the PR. Is that right?

So in this PR, I've included actions that are required to implement the RFC, i.e.
- create the `/rfc` dir
- update the playbook

This makes sense to me in this case because everything affected by this RFC is in the same repo. So the PR contains:
1. the case I made (the RFC itself)
2. the decision made by the org to accept the RFC (the comments)
3. the changes required to implement the RFC (the two actions listed above)

I could definitely see a case for separating (3) out into a separate PR, and I'm happy to do that if you think it makes more sense. OTOH it feels nicely atomic to have it all bundled in one place 🤷 

What do you think @erikdstock ? Which approach is preferable?

admbtlr: To clarify: by "which approach" I mean:

- include the changes required to implement the RFC inside the PR, along with the RFC itself, or
- restrict the PR to only include the RFC, create a new PR for the changes required to implement it

pvinis:
should we have a `_template.md` file in that `RFCs` folder, for easy copy/pasting and editing?

playbooks/rfcs.md
admbtlr: @erikdstock any thoughts?

erikdstock: I think my confusion was around:
a) now RFCs can happen on any repo? _Yes, and peril looks for PRs to a document in the `/rfc` folder or with a title starting with `RFC: `_
b) whether an RFC is accepted or not, what happens to the PR? _it should be merged at resolution with the merge documenting the result_

If these are correct that addresses it for me - I think it would be nice to have some markdown frontmatter showing whether an RFC passed or not.

admbtlr: ping @erikdstock, when you have a moment

erikdstock: ✅ Looks good, thanks!

PR #468, opened 2022-06-27
RFC: use codepush for PR QA and faster releases

pvinis:
codepush


gkartalis (commented):
Love this idea! I am a huge fan of codepush!

One question, using codepush though wouldn't affect the cadence that we do the standard store releases right?

RFCs/codepush-for-mobile-releases.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+### for releases
+
+Any release that includes changes on native code or anything related, like fonts or plist files etc, will still go through the regular appstores process. This is the only way to get these updates to the users' devices, using native updates.
+
+Any release that includes only js/ts file updates, even npm packages (if they don't include native code), they can do either through the regular appstores release process, or codepush. My proposal here would be to have codepush as the default, and only fallback to appstores release process every few releases, or if we have an important change that Apple or Google cares to see. We can keep our cadence as is for now, and re-evaluate later, as we have all the processes in place. But if we want to be very conservative, I would say we can go for a "1 native, 1 codepush" release schedule, unless there are changes that require a native one of course.
```
gkartalis, line 60: I am not sure if we should replace the already existing release process with code push 100% 🤔 

Love the idea of seamless releases that bypass the store checks but not sure if we should do it 100% of the time.

RFCs/codepush-for-mobile-releases.md
```diff
@@ -0,0 +1,75 @@
...
+
+As a result, we end up merging first, and then QA-ing by building a beta. The beta then can be used by non-engineers to try out the changes, and often we need to make another PR to address some issues big or small.
+
+Why not avoid all that, and allow ourselves and everyone else at artsy with the beta app, to try out changes on a PR, without the need to merge first and build a new beta?
+
+- Allow us to do faster releases, more often, and closer to `main`.
+
+They way `force` does releases is a thing to admire. `eigen` is always a bit slow with releases, and for a reason. And that reason is because it's slow to do releases. You see the issue here?
+
+From the second we decide to do a release (after we have QA-ed and everything), a couple of days at least need to pass, for the appstore review. That causes us to be extra careful, because if a bug causes a huge issue, the only way to fix this is by fixing and then waiting two more days, and email Apple to speed this process up. It causes us to be more afraid to break things, because they would remain broken for a long time.
```
gkartalis, line 23: 🔥

joeyAghion:
This is an exciting opportunity. Reviewing PRs would be a nice-to-have, but the ability to ship hot-fixes or save days of release time (and so start to release more often) would be huge. I tend to agree that the slowness and irreversibility of the current process makes us very cautious.

Of course, the devil will be in the details. And adding another dimension (of which release and bundle someone's on) to the already confusing mix of beta/store, staging/production, simulator/device, and feature flag considerations could create even more confusion when it comes to tracing errors.

RFCs/codepush-for-mobile-releases.md
pvinis: it could be a start, and something we try. definitely not replacing everything with codepush just yet.

pvinis:
the most confusing parts right now are not the actual releases, since they are just 2 scripts that work very easily. and codepush would add another script. that part is fine.

the harder part is the scheduling/QAing, which I'm trying to ease with this rfc. so my plan is to not complicate the release docs much more besides the js vs native code, and to uncomplicate by having an easy release/hotfix when we need it.

rajsam003:
Love it ❤️

mdole:
whoa. this sounds like the future, TBH. would be huge if we can just release whenever we want and QA much more easily.

One thing we should get a sense for before committing: how much will it cost? In the [pricing section](https://appcenter.ms/#pricing), I see `240 build minutes per month` and then $40 per month per "build concurrency" and $99 per month per "device concurrency"...I have no clue what that means. Do you understand it @pvinis ?

RFCs/codepush-for-mobile-releases.md
```diff
@@ -0,0 +1,75 @@
...
+
+If we have a way to do hotfixes or even regular feature releases and changes in a matter of seconds or minutes, then it is much much easier to move faster. If something breaks, the fix will be delivered in a few minutes to every user. If we forgot to add something, toggle a flag, change a color, whatever, it can happen in a few minutes.
+
```
admbtlr, line 26: In reality, I don't think this is true.

The only really safe time to check for an updated JS bundle is on app launch (otherwise you risk a mismatch between code and app state). This means that it will take a while for users to actually receive a hotfix, either because (a) the app is running in the background and doesn't get launched for a while, or (b) they're just not launching the app in the first place. 

Meanwhile app store review time is between a few hours and a day, after which time you have a pretty good guarantee (although admittedly not 100%) that users will be updated to the new, hotfixed version. So in reality the improvement that Codepush brings under this scenario is minimal.

RFCs/codepush-for-mobile-releases.md
admbtlr: I agree. I think we should start by reserving it for hotfixes while we evaluate the potential for becoming more CD, and think _very carefully_ about how we would need to change our processes to support that.

pvinis:
@mdole for pricing, it would be free, as we don't need to build with them. my plan is to build in circle, and just deliver it to appcenter only for the OTA part.

I'll verify it's actually free and possible again, but that's how I've used it before.

admbtlr:
I'm very excited about the DX possibilities of Codepush: being able to run arbitrary branches on a device by just selecting a bundle from a dropdown seems like a small miracle to me.

But I'm sceptical about going full CD, partly for the reason that I mentioned in my [comment](https://github.com/artsy/README/pull/468#discussion_r908265997) above. This is a good article about trying out Codepush with a complex app, and how in the end they decided not to use it:

https://www.matthewsessions.com/blog/code-push/

There are a lot of subtle payoffs that we would need to think about before committing to a new set of processes.

But I'm in favour of trying it! I just think we need to be realistic about what we're likely to gain from it.

RFCs/codepush-for-mobile-releases.md
pvinis: I think this is totally the other way.

appstore review times, even if they are fast, do not guarantee a user will get it. the phones usually update be themselves when they are charging. could during the night or in the regular day. some people have the auto update off. there are definitely reasons why the update would not be so immediate.

with codepush, we can check whenever, my proposal would be to check every appstart, probably also every app foregrounding. if a hotfix is out, then it will be downloaded at appstart or fg, and applied on the next start. very easy. also, like I said, we can control that, so if there is a major bug, we can choose to apply it immediately when it's downloaded. yes, it might be jarring to have one flash one time, but better than a huge huge bug. (I would not force the update directly if not huuuge bug.)

there is no issue with app state at all in either case. the app will do migrations as normal, if we have a migration for the update. there is no difference on the apostate if it's a native or a OTA update.

pvinis: > First, [..] it just takes a while for apple/google to review releases and there isn't much we can do about it.

But we can. we can use codepush if we want, to speed things up.

> Second, Apple and Google both don't like shipping regular updates without them knowing about it [..]

Which is why I think its good to do regular appstores updates every few releases. Just to keep things ok for everyone.

> Third, [..] each app launch will result in the app restarting after the new bundle is downloaded, for everyone. [..] especially knowing that the jsbundle is about 8mb now (7.9mb).

I think we are not all on the same page on how this works. Firstly, the size is ok. The regular app on the appstore is 45mb, so an update of 45 and an update of 8 is definitely bigger, if that matters. Secondly, it will not restart the app. As I proposed, the app would download the update, but only apply it on the next appstart, which will not affect the user experience at all.

I think we should start using it for the PR checks first, and once people are familiar on how this works, then we can move to possibly some hotfixes. And from there we can decide if we want to use it for more, or keep it there.

RFCs/codepush-for-mobile-releases.md
```diff
@@ -0,0 +1,75 @@
...
+Use [codepush](https://microsoft.github.io/code-push/) ([github](https://github.com/microsoft/react-native-code-push)) for
+- QA-ing PRs before they are merged, even by non-engineers, and
+- doing faster, leaner releases, more often, and closer to `main`.
+
+## Reasoning
+
+Codepush can be very beneficial to us in two ways.
+
+- Allow us to test/QA PRs before they are merged, and by non-engineers too.
```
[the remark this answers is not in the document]
pvinis: Of course, this is only for js/ts changes. Looking at our past 100 PRs, the vast majority is just js/ts code, which means codepush is a great way to test these. When a PR touches native code, the process stays as is now. We need to merge before we try it on beta.

And yes, we will deploy on every PR on every commit. As we do our tests! The js bundle is built anyway. We will also have one extra step on ci that will actually take that built js bundle and deploy it on codepush, so its available on the betas to run. It is not a big change programmatically, and it is such a huge DX (and non-dev X) improvement to test PRs without first merging.

pvinis:
A little update here.

I want to make this RFC mostly about PR testing first.

I see some people are not too familiar with how things work on codepush, and I see that we definitely need to try out some things first until we are all comfortable with how the whole system works and what control we have over it. So I would like to slightly pivot this RFC for just the PR testing before merging first, and once we are done with that, we can evaluate how/if we want to move further with codepush and use it for hotfixes, and even later if/how we want to use it for releases.

RFCs/codepush-for-mobile-releases.md
MounirDhahri: Agree, let's give it a try. I am excited about it

MounirDhahri: > I think we are not all on the same page on how this works

I understand that we can decide when we want the app bundle to be updated and when we want to make the changes to take effect and this could be at the app start, on AppState change or at the next session, this is clear to me. In most cases I saw though, companies usually tend to do it at the app start since they're trying to make sure users get a fix asap, and if a fix is available, it would wait to get the latest js bundle on app launch and start the app and reloads the splash screen, and if mid-session, it restarts the app if specified (according to code-push docs https://docs.microsoft.com/en-gb/appcenter/distribution/codepush/rn-api-ref#installmode).

If we use code-push for normal app updates, then yes, 8mb is definitely a big advantage compared to 45mb.

> I think we should start using it for the PR checks first, and once people are familiar on how this works, then we can move to possibly some hotfixes

I agree with this, let's try it out. A quick suggestion in this case where we have an RFC for a technology that we might not be all familiar with, is to also add a small paragraph or description for what the tool is and what it does, and if needed how it works. This could help you gather more feedback about it

kathytafel:
So Mounir said Apple "may not like." we need to make sure that whatever we do is within the bounds of the developer agreement we sign, or there is the problem that we could be removed from the store.

> 2.5.2 Apps should be self-contained in their bundles, and may not read or write data outside the designated container area, nor may they download, install, or execute code which introduces or changes features or functionality of the app, including other apps. ...

Certainly we can't add features or change functionality.

pvinis:
I don't know about that.

> 3.3.2 Except as set forth in the next paragraph, an Application may not download or install executable code. Interpreted code may only be used in an Application if all scripts, code and interpreters are packaged in the Application and not downloaded. The only exceptions to the foregoing are scripts and code downloaded and run by Apple's built-in WebKit framework or JavascriptCore, provided that such scripts and code do not change the primary purpose of the Application by providing features or functionality that are inconsistent with the intended and advertised purpose of the Application as submitted to the App Store.

it mentions not changing the "primary purpose" of the app, and the "intended and advertised purpose". Which in my mind we have not changed since I joined artsy. The app allows users to buy and sell art. 🤔

admbtlr:
> The only exceptions to the foregoing are scripts and code downloaded and run by Apple's built-in WebKit framework or JavascriptCore

This is the phrasing they added a while ago to explicitly allow code push; the "or JavascriptCore" part didn't used to be there (see for example [this blog post](https://blog.meteor.com/apple-allows-hot-code-push-in-mobile-apps-c2709c814102) that quotes an earlier version of this paragraph, without the JSC allowance) (remember when Meteor was the future?!). I wonder whether they'll change it soon, now that Hermes is the default runtime instead of JSC?

Anyway, there are plenty of examples of apps that use code push on the app store, so I don't think we should be too worried about that. But I'm still not convinced about the gains we would get from using it for updates, so I agree that we should start by trying it out for PR testing.

kathytafel:
I think we should be looking at the spirit - which is that they want to review what's on the store, and they don't want you changing it after the fact so much. In addition, I think there's an aspect of user consent here. I consent to changing the software on my phone when I download things from the App Store. I personally don't want random bits of software installing itself without my awareness. Obviously for ease of use I don't want to be manually updating everything. However, we shouldn't think that we own the container on everyone's devices. The user owns their device not us. They should choose what goes on it.

For PRs though I think this is effing cool. We should all be able to test what colleagues are adding easily.

kathytafel:
I know all that and it's a matter of when the updates happen - Apple asks me if I want to have it automatically. I don't put google maps on my device so they don't track me. Financial apps mostly put up a forced app update flag to download updates from the store. And to be clear having a client server approach where we can change the server after the fact is adding new functionality. It's just that the user should control when software is installed. Otherwise we are close to malware. Even if we will never ever in 100 million years do something janky with our software, the behavior looks similar. Nobody should trust anyone ever with software. I certainly don't trust WhatsApp or TikTok, for instance, and I long ago deinstalled the Facebook app. Maybe I overindex on threat modeling, but better safe than sorry, especially as we become more of a marketplace.

emmadickson:
I'm on board with this!

brainbicycle:
Thanks for writing this up!

I am on board with trying out codepush for PRs + hot fixes. Not on board with replacing our existing release process with code push. 

There are anecdotal issues quoting App Store rejections explicitly calling out code push all of them after 3.3.2 was updated: https://github.com/microsoft/react-native-code-push/issues/1898, https://github.com/microsoft/react-native-code-push/issues/1297. 
And I agree with Kathy the spirit of the law is Apple's wants to review your code when you add significant functionality. 

The quoted guidelines interpretation is coming from Microsoft who owns codepush so it is important to take it with a grain of salt. It sounds from 3.3.2 that code push is allowed with the callout for JSCore, ...but it also sounds like it is contradicted by 2.5.2. How it is applied is up to Apple and I don't like the idea of banking our continued release of the app on that. Worth mentioning previous companies that provided hot code push functionality assured people that they were in compliance with App Store review guidelines, then one day all those apps started getting rejected: https://news.ycombinator.com/item?id=13817557 The mechanism was different and more risky (Obj-C runtime magic vs js updates) but I worry the risk remains without some specific assurances from Apple.

pvinis:
## Resolution
We decided to do it as a PR List. Everyone agrees that using codepush as a way to test a PR with no native code before merging to main is the best gain for us.

Most people are not into using/replacing our existing release ways. Something to be looked at again in the future.

## Level of Support
1: Overwhelming positive feedback. (for the PR List part.)

## Next Steps
- [ ] I make a tech plan (that I have started) for how we set up the PR list.
- [ ] Get the tech plan approved, and further work continues from there.
- [ ] Implement PR list.

#### Exceptions
N/A

damassi:
As mentioned in Slack, perhaps we can give this a try with Folio to get some experience and then reevaluate in terms of Eigen? A trial run.

PR #473, opened 2022-07-21
RFC: Enable Automerging Feauture for Dependabot PRs

kajatiger:
## Proposal:

We are currently already using dependabot for automatic library version updates. But we still have to merge every one of these updates manually. Let's use the automerging feature from github actions for these dependabot PRs. We could restrict the automerging to minor version updates (or even to only patch updates). And of course we would allow automerging only to the staging environment and only when CI is green. Also this decision would have to be made repo by repo so projects that are not a good fit for the feature (like Eigen) could be excluded. 

Find implementation details in this [tech plan](https://www.notion.so/artsy/Github-actions-allows-dependabot-auto-merge-when-CI-green-45c5213924064657ae2ba5b40c2960d1).

## Reasoning

Most of the dependabot update PRs are trivial and if the CI is green there is no reason to assume that the update would break something. The automerging feature of the github native dependabot is a nice way to reduce the noise for these trivial updates and free up developer’s minds for more interesting work. 
On the other side _not_ merging these updates would hurt us more in the long run and like that we could insure to keep up with the updates more easily.

## Exceptions:

Eigen and other repositories that have requirements for manual backchecking of library updates.

## Additional Context:

You can see our discussion [in slack here](https://artsy.slack.com/archives/C01GNGC4XEZ/p1651077831087519)

## How this RFC is resolved

We'll collect feedback on this RFC from the team for a week. We'll consider this RFC approved
if 50% of the engineering team does not disapprove of this change. Add a 👍 or 👎  reaction to
this proposal to vote.

### DIFF
--- a/playbooks/dependencies.md
+++ b/playbooks/dependencies.md
@@ -78,6 +78,23 @@ The advantages are that this code is unlikely to surprise us in the future (beca
 code) and there are less chances of hitting dependency issues (because it cannot be the dependency of another
 dependency.)
 
+## Automatically updating Dependencies with Dependabot
+
+In most of our repositories we have dependabot enabled (through github). You can find the configurations for the dependabot PRs for a repository in a file called _dependabot.yml_ where things like the package ecosystem and the interval for checking updates are defined. Here is an example:
+```yml
+version: 2
+updates:
+  - package-ecosystem: bundler
+    directory: "/"
+    schedule:
+      interval: daily
+    # Limit to 0 to enable only security updates:
+    open-pull-requests-limit: 0
+    assignees:
+      - fclesio
+    reviewers:
+      - artsy/data-platform
+```
 ## Notes
 
 Prior Artsy engineer’s writing on the topic of dependency management:


iskounen:
I think this is a good idea for reducing manual tasks that carry a low risk. One concern I have is that minor version updates aren't immune to bad changes, as we saw with [colors.js](https://github.com/Marak/colors.js/issues/289) earlier this year. If that is a concern shared by other engineers, we could mitigate it by compiling a list of "trusted" dependencies to start out with.

damassi:
I can only speak to Force, but an enormous number of PRs come through that we mostly just green-light. Maybe @mc-jones can comment more about his process here, but my understanding is that we rarely actually run them, and mostly just merge. For things like colors.js, that was discovered and fixed outside of the dependabot process, and security concerns should be evaluated separately. 👍

mc-jones:
> but my understanding is that we rarely actually run them, and mostly just merge
This is effectively true for minor and patch updates, as between a review of the changelog and test coverage we have a reasonable degree of certainty we will be able to catch most issues. There can be some engineering judgment here, depending on the package, as some can be better at versioning than others. But I'm definitly in favor of automating this for at least patches, and maybe even minor bumps.

joeyAghion:
This weekend's rash of `tzinfo` patches is maybe an interesting counter-example. When a new patch becomes available in a common or low-level dependency (e.g., anything used by rails or other frameworks), PRs would all be opened around the same time, would all be merged around the same time (per this RFC), and would all be released around the same time (per our continuous deployment practice).

It would be preferable if we could roll these changes out more slowly, even when only minor upgrades (just to validate them in production gradually rather than all at once). In today's workflow, that can be accomplished simply by not merging the PRs at the same time.

kajatiger:
Good point @joeyAghion. I know that there is a way to configure dependabot intervals and we could also think about making bigger intervals for fragile/crucial system repos while making the intervals smaller on less impactful services in order to have a bit of a time delay for the dependabot PRs which would also mean that the merge would be not all at once and we would have some time to monitor if anything breaks.

egdbear:
➕  Something we could do more is to add the ignore list and still keep the manual updates on important dependencies. 
One thing which made thinking is to take into consideration the bundle size metric, but that's some other topic.

kajatiger:
I want to add that at my last job we enabled automerging for minor and patch updates and I have only good experiences with that. At least in the ruby ecosystem this lead to a much healthier and up to date codebase with less pain for bigger version upgrades.

PR #474, opened 2022-07-22
docs: Update README.md

patrinoua:
fix broken link

### DIFF
--- a/RFCs/README.md
+++ b/RFCs/README.md
@@ -6,4 +6,5 @@ The RFC is a pull request that creates a new markdown document in this directory
 
 These documents can of course be changed at a later date via a new RFC PR.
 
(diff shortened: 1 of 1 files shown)


RFCs/README.md
```diff
@@ -6,4 +6,5 @@ The RFC is a pull request that creates a new markdown document in this directory
 
 These documents can of course be changed at a later date via a new RFC PR.
 
-For a more detailed description of the process, take a look at our [RFC Playbook](/artsy/README/blob/main/playbooks/rfcs.md).
+For a more detailed description of the process, take a look at our [RFC Playbook](https://github.com/artsy/README/blob/main/playbooks/rfcs.md).
```
admbtlr, line 9: Oops! Thanks for spotting this. But please use `/playbooks/rfcs.md` so that we're not reliant on Github's URL scheme, which in theory could always change (or we could move these docs somewhere else).

PR #477, opened 2022-08-01
feat: minor update after k8s upgrade

artsyjian:
- Update `kubectl` version.
- Datadog links currently hardcode Kubernetes cluster names which change at every Kubernetes upgrade. Let the links refer to `production` and `staging` so we won't have to change them much.


kajatiger:
I tried out the `hokusai configure` with the newer kubectl version and ran a kubectl command and a hokusai command after that and both still worked. I don't know if I need to try out anything else?

artsyjian:
> I tried out the `hokusai configure` with the newer kubectl version and ran a kubectl command and a hokusai command after that and both still worked. I don't know if I need to try out anything else?

thanks for testing. yeah, those should be sufficient.

PR #478, opened 2022-08-09
docs: update commit section of engineering workflow

mc-jones:
This PR updates the commit section of the engineer workflow description. It supports RFC #327 which proposed [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) as Artsy's commit message convention and aims to improve adoption across the engineering team. 

Importantly, it also supports ongoing work to track the ratio of new to rework by parsing through commit messages on Artsy Repos. See [relevant slack thread](https://artsy.slack.com/archives/CA8SANW3W/p1656606603284369).

The list of accepted `types` is based on [commitlint](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional). See [relevant slack thread](https://artsy.slack.com/archives/C02BC3HEJ/p1659976173598959). 

Note: I'll share in #dev before merging, but wanted to be more limited in those who have already commented on this in the first review. But feel free to add other reviewers.

### DIFF
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -54,7 +54,7 @@ but it's, well, nice.
 > A well-crafted git commit message is the best way to communicate context about a change to fellow developers (and
 > indeed to [our] future selves). -[How to write a git commit message](https://chris.beams.io/posts/git-commit/)
 
-In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
+In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We encourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
 
 ```
 <type>[optional scope]: <description>
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -46,7 +46,7 @@ collaboration.
 
 Wait, what? What prefix? We use [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) prefixes in our
 PR titles, because by default, we squash our commits when merging a PR, and this means that the squashed commit
-automatically ends up using the conventional commit format. Using the prefix in your branch name is not essential,
+automatically ends up using the Conventional Commit format. Using the prefix in your branch name is not essential,
 but it's, well, nice.
 
 ### Commit Messages
@@ -54,7 +54,7 @@ but it's, well, nice.
 > A well-crafted git commit message is the best way to communicate context about a change to fellow developers (and
 > indeed to [our] future selves). -[How to write a git commit message](https://chris.beams.io/posts/git-commit/)
 
-In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We encourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
+In support of this, Artsy engineering adopted [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We encourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
 
 ```
 <type>[optional scope]: <description>
@@ -64,9 +64,19 @@ In support of this, Artsy engineering adopted [conventional commits](https://www
 [optional footer(s)]
 ```
 
-Many Artsy engineers will include the Jira ticket id as the `scope`, which will automatically link the PR to the Jira ticket. A basic Artsy commit may look like: 
+Many Artsy engineers will include the Jira ticket id as the `scope`, which will automatically link the PR to the Jira ticket, though this is not required. A basic Artsy commit may look like: 
 
-`feat(FX-4093): curated marketing collections`
+
+```bash
+# commit with type of 'feat' and 'description'
+feat: curated marketing collections
+
+# commit with optional jira ticket number as 'scope'
+feat(FX-4093): curated marketing collections
+
+#commit with optional subject scope
+feat(onboarding): add user onboarding flow analytics tracking
+```
 
 The two main `types` of commits are `fix` and `feat`, however, commits with the following `types` are allowed: 
 - `build`
@@ -83,7 +93,9 @@ The two main `types` of commits are `fix` and `feat`, however, commits with the 
 
 **Engineers should use the `type` that best describes the commit**, but we believe the majority of work will fit within the `feat`, `fix` and (to a lesser extent) `refactor` types.
 
-An additional benefit of using conventional commits is that it allows us to track the ratio of new work (i.e., feature work) against rework (i.e., bug/regression fixes). We currently use this ratio as one measure of code quality. **Importantly, this means that consistent commit messages and type accuracy supports meaningful metrics**. The ratio of feature to rework can be viewed [here](https://artsy.net/rework-metric).
+In instances where a PR will be simply merged - as opposed to being squashed before being merged - we generally expect each individual commit within the PR to follow the Conventional Commit format. **However, a developer may choose to use a good but "*unconvenitonal*" commit messages within a PR instead if they feel it will result in a more precise commit history on the `main` branch** (relevent discussion in this [PR](https://github.com/artsy/README/pull/478)). The PR title should still use the Conventional Commit format (see the [Pull requests](#pull-requests) section).
+
+An additional benefit of using Conventional Commits is that it allows us to track the ratio of new work (i.e., feature work) against rework (i.e., bug/regression fixes). We currently use this ratio as one measure of code quality. **Importantly, this means that consistent commit messages and type accuracy supports meaningful metrics**. The ratio of feature to rework can be viewed [here](https://artsy.net/rework-metric).
 
 ### Pull requests
 
@@ -108,7 +120,7 @@ Include any details that add context to the PR, as
 As mentioned above, PRs will be squashed when merged, unless there's a good reason not to (see the [RFC](https://github.com/artsy/README/issues/327)
 that initiated this practice for more explanation and discussion). When they're squashed the title of the PR will
 be used for the single squashed commit, so make sure that you use [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/)
-format in your PR title.
+format in your PR title. 
 
 #### Assignees and Reviewers
 
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -54,7 +54,7 @@ but it's, well, nice.
 > A well-crafted git commit message is the best way to communicate context about a change to fellow developers (and
 > indeed to [our] future selves). -[How to write a git commit message](https://chris.beams.io/posts/git-commit/)
 
-In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://excalidraw.com/#json=dMGWGEi9PAmxEZXBsC0kG,kUHekoAEfwndFljUuTUqOA)): 
+In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
 
 ```
 <type>[optional scope]: <description>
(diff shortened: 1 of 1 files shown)


playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,38 @@ PR titles, because by default, we squash our commits when merging a PR, and this
 automatically ends up using the conventional commit format. Using the prefix in your branch name is not essential,
 but it's, well, nice.
 
-### Commits
+### Commit Messages
 
 > A well-crafted git commit message is the best way to communicate context about a change to fellow developers (and
 > indeed to [our] future selves). -[How to write a git commit message](https://chris.beams.io/posts/git-commit/)
```
mc-jones, line 55: I like the quote, but this may be confusing to link to this article that doesn't follow the convention?

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,38 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
+- `perf`
+- `refactor`
+- `revert`
+- `style`
+- `style`
+- `test`
+
+**Engineers should use the `type` that best describes the commit**, but we believe the majority of work will fit within the `feat`, `fix` and (to a lesser extent) `refactor` types.
+
+An additional benefit of using conventional commits is that it allows us to track the ratio of new work (i.e., feature work) against rework (i.e., bug/regression fixes). We currently use this ratio as one measure of code quality. **Importantly, this means that consistent commit messages and type accuracy supports meaningful metrics**. The ratio of feature to rework can be viewed here (TODO: link to public view once done).
```
mc-jones, line 83: @joeyAghion - I saw that you were working on making the view public, but let me know if I shouldn't wait for that.

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,38 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
+- `chore`
+- `ci`
+- `docs`
+- `feat`
+- `fix`
+- `perf`
+- `refactor`
+- `revert`
+- `style`
+- `style`
```
MounirDhahri, line 78: **suggestion:**
```suggestion
- `style`
```

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,38 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
+The two main `types` of commits are `fix` and `feat`, however, commits with the following `types` are allowed: 
+- `build`
+- `chore`
+- `ci`
+- `docs`
+- `feat`
+- `fix`
+- `perf`
+- `refactor`
+- `revert`
```
MounirDhahri, line 76: **issue(non-blocker):** `revert` is still not yet supported in our danger rules, we should probably add it here https://github.com/artsy/peril-settings/blob/1d4504a58466d0a250a75cb30f73827c8d3c2063/org/allPRs.ts#L225

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,38 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
-### Commits
+### Commit Messages
 
 > A well-crafted git commit message is the best way to communicate context about a change to fellow developers (and
 > indeed to [our] future selves). -[How to write a git commit message](https://chris.beams.io/posts/git-commit/)
 
-The code change itself should be internally consistent and coherent. If you're modifying a method signature,
-propagate the change to calling code. If you've implemented a new behavior, update test assertions to match. Except
-in rare cases where you intend to share a work-in-progress, commits should not leave tests in a failing state.
+In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://excalidraw.com/#json=dMGWGEi9PAmxEZXBsC0kG,kUHekoAEfwndFljUuTUqOA)):
```
mdole, line 57: wrong link 😉 

```suggestion
In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
```

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,38 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
-propagate the change to calling code. If you've implemented a new behavior, update test assertions to match. Except
-in rare cases where you intend to share a work-in-progress, commits should not leave tests in a failing state.
+In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://excalidraw.com/#json=dMGWGEi9PAmxEZXBsC0kG,kUHekoAEfwndFljUuTUqOA)): 
 
-Commit subjects should concisely describe the enhancement or fix in the imperative form ("Add tracking of email
-deliveries" or "Undo home page A/B test"). Git subcommands (like `git blame`, `revert`, `log`...) expect this
-format and become more transparent to use when you adhere to that convention.
+```
+<type>[optional scope]: <description>
```
mdole, line 60: **suggestion:** some folks at Artsy include the Jira ticket number in their PR title (which, of course, ends up being the squashed commit) because we have things (Peril, I think?) set up to automatically perform the issue linking. Could we mention and codify this? Might actually belong in the PRs section below, but I think it would be nice to call out.

Some recent examples from me:

```
feat(FX-4093): curated marketing collections

chore[FX-3378]: deprecate addToCalendar helper function

chore[FX-3378]: stop using Cohesion addToCalendar helper
```

mdole (approved):
thanks for putting this together @mc-jones! one fix and a suggestion for ya :)

Mitch-Henson (approved):
I think this is great and I love the standardisation for PR messages. Going through git blame and finding old PRs which clearly link to tickets and give a good description has been super useful.

To confirm, this only applies to PRs right? I remember a conversation within a slack thread (that I can't find with a cursory search) where this seemed to be a hot topic. If we are squashing all commits when merging then I guess only the PR title is relevant

playbooks/engineer-workflow.md
joeyAghion: Yes but I got stuck on a Looker thing. Maybe link to what we have now, or just a placeholder (internal Notion page or artsy.net shortcut?) that we can eventually update with the URL.

playbooks/engineer-workflow.md
mc-jones: Yeah, that's a good call! I will add an Artsy-specific example and include that the ticket id can be used included in the `optional scope`, which will create the automatic jira link.

playbooks/engineer-workflow.md
mc-jones: Good catch. Put up a PR: https://github.com/artsy/peril-settings/pull/198

mc-jones:
@Mitch-Henson 

> To confirm, this only applies to PRs right? I remember a conversation within a slack thread (that I can't find with a cursory search) where this seemed to be a hot topic. If we are squashing all commits when merging then I guess only the PR title is relevant

The measure looks at all commits on the `main` branch. So if you `squash` it only applies to the PR. But if you do a pure merge then all commits will be counted. So if a straight merge is the preferred merge method, the submitter and reviewers/assignee should make sure each commit in the PR follows the convention. 

Was the slack thread you're thinking of [this one](https://artsy.slack.com/archives/CA8SANW3W/p1660079290118909) between @damassi and I taking about Force?

playbooks/engineer-workflow.md
mc-jones: Ok. Added an artsy shortcut to the looker page via https://artsy.net/rework-metric.

Mitch-Henson:
@mc-jones that wasn't the slack thread I was thinking of. I was referring to one at least a few months old but is still in my mind. I think @jonallured was in it and there were very strong opinions for both points of view. I was impressed how polite the discussion was for how strong the opinions were.

The TL;DR of the slack thread I was mentioning was that as long as we don't dictate how each single commit needs to be structured (anecdotally most PRs as we use squash), then only the PR structure is relevant which I agree with. So totally happy for this change to go ahead as I believe that PR's should be standardised anyway

MounirDhahri (approved):

jonallured:
I'm confused about this PR because it's labeled as a `feat` but reads like an RFC. I strongly disagree with this change and would encourage us to not merge it without due discussion!

I have no problem with this structure at the PR level but I believe this is too prescriptive at the commit level. Steve Hick wrote a nice blog post about this very thing here:

https://artsy.github.io/blog/2020/08/11/improve-pull-requests-by-including-valuable-context/

And I'll take the PR of mine that he called out as doing this well:

https://github.com/artsy/convection/pull/645

Here's the log of changes:

```
* 1d9afdf Offer params should override defaults
* b52c5d8 Extract locals around creating an offer
* d2b7831 Do some reorganization of spec file
* 4316914 Simplify specs when there is a PS
* 91eba9c Use the rspec change matcher when possible
* d793c44 Simplify error assertions
* a10a54e Only use reload when you have to
* 756311f Run prettier
* ec4fa9b Use consts for offer states
* 94a5d20 Use lazy load unless you can't
* a66e684 Move to using consts instead of strings
* 0875eaa Promote submission fabrication, leave state
* 1bd026c Use more generic name
* 605eb73 Push fabrication down
* 340e37d Extract local
* 7c81e45 Forgot one
```

If this work had followed the commit-level rules of this PR then it would look a lot different. I _think_ we'd leave off the scope of each commit (right?) but I'd be asked to include the type? So maybe that looks like this:

```
* 1d9afdf feat: Offer params should override defaults
* b52c5d8 refactor: Extract locals around creating an offer
* d2b7831 refactor: Do some reorganization of spec file
* 4316914 refactor: Simplify specs when there is a PS
* 91eba9c refactor: Use the rspec change matcher when possible
* d793c44 refactor: Simplify error assertions
* a10a54e refactor: Only use reload when you have to
* 756311f chore: Run prettier
* ec4fa9b refactor: Use consts for offer states
* 94a5d20 refactor: Use lazy load unless you can't
* a66e684 refactor: Move to using consts instead of strings
* 0875eaa refactor: Promote submission fabrication, leave state
* 1bd026c refactor: Use more generic name
* 605eb73 refactor: Push fabrication down
* 340e37d refactor: Extract local
* 7c81e45 fix: Forgot one
```

I don't think this is better in terms of an engineer reading the history to discover how things have changed over time.

Does this format help with the analysis of our work? Does this tagging ensure the ratio comes out right? Maybe? But why not point our analysis at PRs rather than commits! I'm proud of the work I did there to create a history that my PR reviewers could read and digest and I believe it would be a shame if our docs directed folks away from this type of approach.

admbtlr:
My understanding here is that we're only interested in commits that end up in `main`. So if you're squashing commits on merge, then it's only the PR title (or manually entered commit message) that matters, not the individual commits. If you're _not_ squashing, then yes you will need to prepend all your individual commits, as we've discussed in the past.

This isn't any different to what we're already doing, is it? Or am I missing something @jonallured ?

mc-jones:
@admbtlr - @jonallured and I chatted quickly last week on this and I'm going to amend this a bit. Specifically, I'm going to include a line to clarify that, in instances where developers feel it will result in a more precise commit history to _not_ use a conventional commit message for a specific commit that will be merged into `main`, they may do so (this doesn't apply to PR titles).

The goal of this is to continue to encourage conventional commits as our standard, but allow for some flexibility to use a good, but not spec-adhering, commit where it may not be the most appropriate.

mc-jones:
Also, following my own suggestions, I'm going to change the `type` from `feat` to `docs`.

damassi:
The issue @jonallured is that this PR was mostly likely not squashed, meaning each of those commits run the risk of not being measurable by our system. And lots of PRs are only one or two commits, so adding the appropriate tag is pretty trivial. 

Secondly, it shouldn't be up to the reviewer to remember to squash the commit; that's asking too much of the process. So its either: dev makes nice orderly commit history in the style that they like -- and then either rebases to squash the commit before merge, or instructs the reviewer to only "approve" the PR so that the dev can update github UI before merge. 

All of this complexity could be alleviated by tagging the commit with what it was. Sure, every one of those commits was a "refactor", but is this representative of a larger body of work? And additionally, if the history *wasn't* squashed, and one of those `refactor` commits introduced a bug (as refactor commits usually do) how is another person going to come along during a git bisect and know where to start? There's no indication that such and such commit in your example is a refactor, leading to wasted time.

jonallured:
> meaning each of those commits was not measurable by our system

Right @damassi which is why I'm saying we should measure at the PR level NOT the commit level. I think this is the direction you're headed, right @mc-jones?

> The goal of this is to continue to encourage conventional commits as our standard, but allow for some flexibility to use a good, but not spec-adhering, commit where it may not be the most appropriate.

I support this goal and appreciate the flexibility! 🙌

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,41 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
-### Commits
+### Commit Messages
 
 > A well-crafted git commit message is the best way to communicate context about a change to fellow developers (and
 > indeed to [our] future selves). -[How to write a git commit message](https://chris.beams.io/posts/git-commit/)
 
-The code change itself should be internally consistent and coherent. If you're modifying a method signature,
-propagate the change to calling code. If you've implemented a new behavior, update test assertions to match. Except
-in rare cases where you intend to share a work-in-progress, commits should not leave tests in a failing state.
+In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We ecourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)):
```
pvinis, line 57:
```suggestion
In support of this, Artsy engineering adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) as its commit message convention (see RFC [Best Practices for Naming and Merging PRs](https://github.com/artsy/README/issues/327)). We encourage you to read the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) for more detail but, in general, all commits should match the following format ([see examples](https://www.conventionalcommits.org/en/v1.0.0/#examples)): 
```

playbooks/engineer-workflow.md
```diff
@@ -49,26 +49,41 @@ PR titles, because by default, we squash our commits when merging a PR, and this
...
+- `chore`
+- `ci`
+- `docs`
+- `feat`
+- `fix`
+- `perf`
+- `refactor`
+- `revert`
+- `style`
+- `test`
```
pvinis, line 82: about the tooling that we currently have, how adjustable is it? if we wanted to start using a new type, say `potato`, for every PR that includes the a gif of a potato in a repo.
how easy/hard is that adjustment?

mc-jones: This should just affect Peril and commitlint. For Peril, we just need to add it to the `danger` plugin. We did this [recently](https://github.com/artsy/peril-settings/pull/198) for `refactor`. For `commitlint`, looks like we would just need to add a rule in the config (I believe it's currently only in [Forque](https://github.com/search?q=org%3Aartsy+commitlint&type=code)). Not sure if there is more tooling I'm unaware of?

mc-jones:
> I think this is the direction you're headed, right @mc-jones?

I'm happy with this path as I believe it is the best tradeoff to achieve the goal of allowing us to meaningfully evaluate the commits without restricting developers' ability to make judgments about when valuable context would be lost by forcing the convention. 

I also want to emphasize that part of this effort is to use the data to better inform our practices. Once we share this, if we are finding that there are still a large number of unconventional commits, we can relook at this with actual data points.

egdbear (approved):
🚀

PR #487, opened 2022-12-09
feat: clarify significant staging issues should be treated as incidents

mc-jones:
This PR is a follow-up to incident [161](https://artsy.slack.com/archives/C04BPTCCZ9N). In the review discussion, we agreed that it makes sense that incidents affecting staging, but not production, still go through the same process. This just adds instruction to that effect in the incident playbook. 

If everyone is on board I'll share the change in #dev, as a follow-up. 

Note: we had discussed in the call that we keep the discussion to #dev, and not use #incidents when it just affects staging. While finalizing the postmortem, It seemed overly confusing to have two different processes. So, instead, I just added a line here to emphasize the scope of the incident should be communicated. 

Happy to change course if anyone feels this was a bad call. 

[Platform-4787]

### DIFF
--- a/playbooks/incident-handling.md
+++ b/playbooks/incident-handling.md
@@ -315,6 +315,8 @@ Many people follow the #incidents channel closely, so avoid noise by keeping dis
   <!-- TODO: define how to apply priority levels more consistently -->
 - Take your best guess at the other information; it can always be edited later.
 
+Issues isolated to staging (i.e., issues that do not have have an observable impact on production) resulting in major workflow disruptions should also be treated as incidents. When communicating a staging incident, it should be made clear that there is no preceived impact on production to avoid creating confusion for the team members following the incident.
+
 2. **Create a Slack channel** dedicated to the incident's technical resolution
 
 - Follow the **Create Slack channel** link from the OpsGenie incident page and create a `#inc-*` channel named with


dzucconi (approved):

kajatiger (approved):

playbooks/incident-handling.md
```diff
@@ -315,6 +315,8 @@ Many people follow the #incidents channel closely, so avoid noise by keeping dis
   <!-- TODO: define how to apply priority levels more consistently -->
 - Take your best guess at the other information; it can always be edited later.
 
+Issues isolated to staging (i.e., issues that do not have have an observable impact on production) resulting in major workflow disruptions should also be treated as incidents. When communicating a staging incident, it should be made clear that there is no preceived impact on production to avoid creating confusion for the team members following the incident.
```
joeyAghion, line 318: "Staging" is just one of our development tools, and I think this could apply equally to other tools (e.g. jenkins), so I'd consider describing these more generally as "non-production" disruptions that significantly affect teams' ability to function.

I would also recommend explicitly recommending the "low" or "informational" severity level for these.

ovasdi (approved):
👍 

**Suggestion**:
- putting the paragraph under [responding-to-other-incidents](https://github.com/artsy/README/blob/main/playbooks/incident-handling.md#responding-to-other-incidents) as opposed to _Raise_ section.
- mention that Staging incidents should initially be raised in #dev (and escalated to #incidents if deemed to be affecting production)
- clarify what "treated as incidents" means exactly.
> "...disruptions should also be treated as incidents (opsenie incident created, timeline tracked, postmortem added, incident review scheduled)"

**Super minor** : While you're in there fix the number list under Raise. I noticed that in the [Raise](https://github.com/artsy/README/blob/main/playbooks/incident-handling.md#2-raise) section the numbered items skip one. It goes from 2 to 4.

PR #491, opened 2022-12-19
docs: Update rfcs.md

patrinoua:


patrinoua:
Actually not sure this is the right way to add a new RFC. This is the old way right? 
Where do we add the new PR when creating an RFC?

PR #501, opened 2023-02-10
fix: add Python as it has been in use in various repos.

artsyjian:
In recent discussions around an [Opstools PR](https://github.com/artsy/opstools/pull/12), we noticed that the Tech Radar has no entry for Python. It seems we should add such an entry, because there are significant [repositories with Python code](https://github.com/orgs/artsy/repositories?language=python&q=&sort=&type=all), Hokusai/Motion/Zenith being the main ones.

The `description` field for the entry is left blank, to be filled in based on discussion in this PR, therefore, the PR is tagged `Draft`.


cavvia (approved):
Good call. I need to add Airflow and AWS Batch as platforms as well, they appear to be missing.

joeyAghion:
We should definitely clarify its "adopt" status for data purposes. I like the description @cavvia suggests.

For infrastructure tooling (starting with `hokusai` and `opstools`/`ops-util`), it feels more incidental. I think @izakp chose it since it was most familiar and had strong library support, which are the same motivations the org _usually_ has around ruby. I'd argue we have more broad internal knowledge of maintaining and deploying ruby as well, so I'd want that to be considered before starting new python projects in non-data areas.

artsyjian:
The description field has been filled in [as suggested](https://github.com/artsy/README/pull/501#pullrequestreview-1293644602).

PR #502, opened 2023-02-10
doc: clarify ruby's adopt status

joeyAghion:
Inspired by https://github.com/artsy/README/pull/501, let's clarify ruby's "adopt" status, especially for data-backed APIs and services.

We have many ruby back-ends and significant built-up experience maintaining and evolving them. (The only exceptions I can think of are Positron, Causality, and APRD, which are unmaintained or deeply out-of-favor.) It should be the default option for any new back-end needs.


artsyjian (approved):
agreed

PR #503, opened 2023-02-16
Suggest README language around setup scripts

joeyAghion:
As discussed, let's consistently advise _reviewing_ setup scripts, and running them or the _equivalent_ steps for one's environment.


erikdstock (approved):
Thanks for this!

PR #508, opened 2023-03-22
feat: mention hokusai update-config flag for deployment rollback

artsyjian:
`hokusai production deploy <tag>` updates Kubernetes `Deployment` resources only, which is insufficient for  
[Incident 170](https://artsy.slack.com/archives/C04RN72NARG) where we have to revert `Ingress` (in addition to `Deployment`).

Let the doc mention `--update-config` which can be used to update all Kubernetes resources defined in `production.yml` spec.

### DIFF
--- a/playbooks/deployments.md
+++ b/playbooks/deployments.md
@@ -90,9 +90,25 @@ README.
 
 ## "Hot" fixes and roll-backs
 
-Sometimes there are urgent fixes to release. We always prefer using the typical pipeline to "roll forward," even in
-cases of reverted PRs or timely fixes. When it's critical to avoid the delay of a full CI run or staging deploy,
-it's possible to simply `hokusai production deploy <tag>` where _tag_ is an image tag (including git SHAs) and can
-refer to a previous release such as `production--2018-09-25--10-19-2`. (Run `hokusai registry images` to see recent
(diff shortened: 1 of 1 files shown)


playbooks/deployments.md
```diff
@@ -90,9 +90,25 @@ README.
...
+- From the list, identify the last good working image - the one you want to revert to. Note it's tag.
+
+- Revert Production deployments to the last good working image.
+  ```
+  hokusai production deploy <tag>
+  ```
+
+  That updates Kubernetes `Deployments` only. If other Kubernetes resources (e.g. `Ingress`) have to be reverted also, add `update-config` flag:
+  ```
+  hokusai production deploy <tag> --update-config
```
joeyAghion, line 111: Any reason not to _always_ advise this additional argument? I think it will be a no-op when the spec is unchanged.

joeyAghion (approved):

playbooks/deployments.md
artsyjian: to allow the possibility of rolling back or hotfix only the code, when a troubled release had both code and k8s spec changes. probably unlikely situation, but seems we can't rule it out, as it all depends on what problem there is and how we decide to fix.

PR #515, opened 2023-08-02
Add RFC Template

jonallured:
Unless I'm missing something our issue template for RFCs does not work anymore. I searched for GH docs and it appears they want these templates in `.github/ISSUE_TEMPLATE/` so I used their builder to create what you see here. I just copy/pasted from the `_templates/rfc.md` file so it should be a match.


damassi:
👍

damassi (approved):

PR #517, opened 2023-08-03
Update practice leads

jonallured:
While reviewing all the stale info in this repo I decided to start with updating this list.

/cc @cavvia @tam-kis @TMcMeans @MrSltun @gkartalis @mc-jones


gkartalis (approved):
thanks for updating!

PR #518, opened 2023-08-03
Improve clarity around rfc process

jonallured:
I was very confused yesterday about the current RFC process. The last time I made an RFC it was an issue on this repo so I assumed that was still the way. Then I found in the root README of this project a link to create an issue so I clicked it. I expected to see an issue template populate but it didn't so then I was confused. I asked on Slack here:

https://artsy.slack.com/archives/CP9P4KR35/p1691010929981279

And the hive mind steered me towards re-creating the template. This was incorrect.

The next day my pal @lidimayra noticed my mistakes and confusion and linked me to the playbook and the RFC where we changed course - RFCs should be authored as markdown files in the `/RFCs` folder and then a PR should be opened to discuss them. I had completely missed this development!

So then I thought the right thing to do would be to come back to this repo and cleanup things that caused me confusion to hopefully save the next person from making my same mistakes.

* remove the issue template that I added yesterday
* update the root README to point people to the playbook
* update the README in the `/RFCs` in the same way
* relocate the templates to the `/RFCs` folder so that everything is in one place
* update the playbook

I think this is better and should result in less confusion but it's probably not perfect - what did I miss?


lidimayra (approved):
oh wow!!! I had no idea there would be so many places to be updated about this!!
Thanks for taking care of it, the current process is much clearer now!!

damassi (approved):
Thank you @jonallured!

PR #519, opened 2023-08-03
Add a note about OpsGenie notifications to the "Track the incident" section

dariakoko:
Addressing action item from inc-180

https://artsy.app.opsgenie.com/reports/post-mortem/bc69f1ee-85d1-4a0d-b956-a1079e8645cb/detail

### DIFF
--- a/playbooks/incident-handling.md
+++ b/playbooks/incident-handling.md
@@ -310,6 +310,10 @@ Issues not observably effecting production, but resulting in major workflow disr
 
 1. **Track the incident** by creating an incident in [OpsGenie](https://artsy.app.opsgenie.com/incident/list) based
    on the available information
+   
+> After you submit the incident, another onCall engineer will be notified via an alert from OpsGenie
+> 
+> Add responders immediately if the failure requires attention urgently and you cannot handle it yourself, the responder will be notified via an alert from OpsGenie as well
 
 - Select _Artsy_ as the impacted service.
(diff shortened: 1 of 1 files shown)


playbooks/incident-handling.md
```diff
@@ -310,6 +310,10 @@ Issues not observably effecting production, but resulting in major workflow disr
 
 1. **Track the incident** by creating an incident in [OpsGenie](https://artsy.app.opsgenie.com/incident/list) based
    on the available information
+   
+> After you submit the incident, another onCall engineer will be notified via an alert from OpsGenie
+> 
+> Add responders immediately if the failure requires attention urgently and you cannot handle it yourself, the responder will be notified via an alert from OpsGenie as well
```
ovasdi, line 316:
```suggestion
> Add responders to the OpsGenie incident if the issue requires urgent attention and you cannot diagnose/address/resolve the incident yourself. Once added, the responders will be notified by OpsGenie
```

ovasdi (commented):
Thanks for following up so quickly on this AI. Added a suggestion for how to word these _tips_ in another way.

The _Raise_ section describes how to generate an OpsGenie incident so I wonder if it would be more appropriate to instead edit/update the [Escalate](https://github.com/artsy/README/pull/519/files#diff-415fc7b74e5e5282cbbb89bfb542da0213130461d541605d2c776fedcb55a4ebR380) section as opposed to adding that info under _Raise_?

PR #525, opened 2023-10-02
docs(tech-radar): Add Next.js to the list

damassi:
Added `adopt` to next, and the word "should" in the description! Thoughts?

### DIFF
--- a/playbooks/technology_radar/artsy-tech-radar.csv
+++ b/playbooks/technology_radar/artsy-tech-radar.csv
@@ -1,6 +1,7 @@
 name,ring,quadrant,isNew,description
 GraphQL,adopt,languages & frameworks,TRUE,"Appreciated for its flexibility and efficiency for clients."
 JSON,adopt,languages & frameworks,FALSE,"We prefer JSON over most other data formats for its self-describing properties."
+Next.js,adopt,languages & frameworks,FALSE,"New apps with frontends that are GraphQL centric should use Next. See Forque, Volt 2."
 Python,adopt,languages & frameworks,FALSE,"For ETL, Analytics and ML workloads we use Python as a data engineering standard."
 Rails,adopt,languages & frameworks,FALSE,
 React Native,adopt,languages & frameworks,FALSE,"New iOS work should use React Native. See also Objective-C."


jonallured (approved):
Agreed!

anandaroop (approved):

joeyAghion:
Thanks for adding this, as it _is_ a major stack change to document.

I actually think of Next as borderline between Trial and Adopt. We're trial-ing it in the _right way_, in my opinion: we built an internal app, then we built a partner-facing app, taking those production [but low-risk] lessons and propagating them. We haven't yet used it for our highest-volume products. The risks or questions I'd hope are addressed before that include:
* how stable the framework is (e.g., do upgrades require a lot of work?)
* the ENV-handling incompatibility in v>=13 (possibly addressed in [volt-v2 #760](https://github.com/artsy/volt-v2/pull/760) and [#752](https://github.com/artsy/volt-v2/pull/752))
* the reports of client-side instability (unresponsive tabs)

Web practice, or the devs working in these repos, can decide this Trial vs. Adopt question as far as I'm concerned. I just think it's worth that explicit feedback about how it's gone so far. (And Next only has to be better than the alternative, not _perfect_.) Has this happened? Should I raise it? Happy to merge if so.

damassi:
To answer:

> How stable the framework is (e.g., do upgrades require a lot of work?)
Usually none. (Though going from `pages` to `layouts` is a major change)

> the ENV-handling incompatibility in v>=13 (possibly addressed in https://github.com/artsy/volt-v2/pull/760 and https://github.com/artsy/volt-v2/pull/752)
This is done! Though we should get it over in Forque as well and let things simmer. 

> The reports of client-side instability (unresponsive tabs)
This is specific to Convos. Addressed change to fix this; and my take is this isn't a next thing but rather how we architected our page. Though that said, there was a ping about this today which i suspect is a page cache issue. We will see. 

>Web practice, or the devs working in these repos, can decide this Trial vs. Adopt question as far as I'm concerned. I just think it's worth that explicit feedback about how it's gone so far. 

I think this goes without saying? Every web dev (everywhere) wants to work with next, and every team member who has worked with our next apps has described it as nothing less than a delight. It is the standard for Javascript/Web frameworks, has the best web devs in the industry on the team (including all of the former facebook react core team engineers), is free, etc, etc. What more do we need?

joeyAghion:
> I think this goes without saying?

This is sort of the point of the radar... to take things that might go without saying, or be products of inertia, and instead make them explicit milestones, with discussion and even version control of the state changes. _I_ learned something, so I'm glad I shared my concerns.

Given that 3 of the most active Forque contributors have already weighed in affirmatively here, I'll go ahead and merge.

Question for later: you mention front-ends specifically in this entry, and I understand Next is focused on React. If we were rebuilding a node system without a front-end (like Metaphysics), would we still use express today?

damassi:
> would we still use express today?

Yes indeed. There are other, newer options, but we shouldn't deviate far from the mean for something as critical as a backend server framework. Express is still by far #1. If we were to choose something newer and more fancy there would be a lot more risk for little gain.

PR #530, opened 2023-11-06
doc: update preferred S3 path for shared configuration

joeyAghion:
Back in https://github.com/artsy/README/pull/368, I proposed a convention for storing shared configuration for development environments in S3. That convention has been working, but I regret choosing the particular path (`artsy-citadel/dev/.env.<project>`) for a few reasons:

* It requires renaming the file as part of the `aws s3 mv` operation. This misses an opportunity for the file to self-document how it should ultimately be named. It also creates confusion (when it's copied down with its original name, or when a `.env.shared` is mistakenly copied up to S3).
* It doesn't accommodate multiple files, which a few projects (Vortex, Eigen, Infrastructure, Energy...) have needed.

Instead, I think we should source shared config from paths like `artsy-citadel/<project>/.env.shared`. Projects that depend on multiple files have already started using that pattern, and moving the existing (<40) files should be easy, like:

```bash
aws s3 cp s3://artsy-citadel/dev/.env.admin_metadata s3://artsy-citadel/admin_metadata/.env.shared
```

Then each project's setup script would need an update to pull from the new location (also easy, but tedious). Finally, the files under `dev/*` can be cleaned up to minimize confusion.

I'll start this process, but wanted to solicit feedback since it's a slight change to our playbook.

### DIFF
--- a/playbooks/development-environments.md
+++ b/playbooks/development-environments.md
@@ -91,7 +91,7 @@ Docker-based development respects these files by listing both in docker's `env_f
 To _update_ shared configuration values, simply modify `.env.shared` and re-upload it to its shared location as
 `.env.<project>`. E.g.:
 
-    aws s3 cp .env.shared s3://artsy-citadel/dev/.env.zulu
+    aws s3 cp .env.shared s3://artsy-citadel/<project>/.env.shared
 
 If you are _adding_ a new configuration value and need it reflected on remote
 environments, check this


mc-jones (approved):
I like this pattern! It will be nice to have it consistent with how we have engineered `vault` to [work](https://www.notion.so/artsy/Hashicorp-Vault-developer-instructions-77d94af51f714d51bb44049f4f2027bc?pvs=4#c478193b2a9747459310a0f85a852cca) (assuming we move forward with it). 

These may already be on your radar but looks like there may be some changes required outside the `setup` scripts looking at this [search](https://github.com/search?q=org%3Aartsy+artsy-citadel%2Fdev&type=code&p=1). For example:
- `Motion`'s [makefile](https://github.com/artsy/motion/blob/cfb7ed839ac6e6c393879166216f43b92e0a3fd0/Makefile#L18)
- `Gravity`'s [terraform] permissions [config](https://github.com/artsy/infrastructure/blob/fdfc66d77abf7d00554225d33b1077ea6563b0dd/terraform/shared/iam_policies.tf#L818)

joeyAghion:
I actually did a similar search, but wasn't sure about [this one](https://github.com/artsy/infrastructure/blob/fdfc66d77abf7d00554225d33b1077ea6563b0dd/terraform/shared/iam_policies.tf#L818). Do you know why the gravity IAM profile needs access to the development configuration?

mc-jones:
> I actually did a similar search, but wasn't sure about [this one](https://github.com/artsy/infrastructure/blob/fdfc66d77abf7d00554225d33b1077ea6563b0dd/terraform/shared/iam_policies.tf#L818). Do you know why the gravity IAM profile needs access to the development configuration?

Tracked down to https://github.com/artsy/infrastructure/pull/464. I _think_ it may have been needed for a jenkins job.

joeyAghion:
Since k8s crons and Batch jobs get their config in other ways these days, I'm inclined to drop that `Allow` block. Let me know if you think otherwise.

PR #533, opened 2024-02-08
doc: link and expand open-source guidelines

joeyAghion:
Minor additions to:
* Link our open-sourcing playbook to the other playbook's guidelines for public vs. private content
* Mention the additional case that came up with respect to Volt-v2, of code repositories that aren't necessarily sensitive themselves but benefit from the safety of private discussion


damassi (approved):
👍

jonallured (approved):

PR #537, opened 2024-04-02
chore: remove peer learning group section

anandaroop:
Just keeping our public version of this in sync with [the change](https://artsy.slack.com/archives/C02BC3HEJ/p1711989001006189) we made in Notion


artsyjian (approved):
good catch!

PR #544, opened 2025-02-25
chore: senior staff ladder draft

pam-:
This PR aims to open up the discussion further regarding the next level of Staff at Artsy.

Open to comments/pushback/etc.

### DIFF
--- a/careers/ladder.md
+++ b/careers/ladder.md
@@ -107,6 +107,7 @@ To help with the visualisation of the ladder we have created this visual, inspir
       <td class="border-top">Senior Engineer 1</td>
       <td class="border-top">Senior Engineer 2</td>
       <td class="border-top">Staff Engineer</td>
+      <td class="border-top">Senior Staff Engineer</td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


careers/ladder.md
```diff
@@ -421,6 +458,13 @@ To help with the visualisation of the ladder we have created this visual, inspir
           <li>Can present larger technical initiatives to leadership.</li>
         </ul>
       </td>
+      <td>
+        <ul>
+          <li>Cultivates strong relationships with all disciplines and managers within PDDE and PDDE leads and prioritizes efforts based on their needs and the overall product strategy.</li>
```
luizonthedrums, line 463: PDDE leads needs? What's an example of this currently?

joeyAghion:
I believe our expectations for the "terminal" levels of senior and staff are already quite high and our team is productively challenged to meet them. There certainly _is_ potential for even more impact, but it feels premature to extend our ladder this way. I'd rather strengthen and refine how we apply the existing expectations.

I'm sure we'll revisit this and borrow from this PR when we do, but closing for now.

PR #546, opened 2025-06-24
chore: update slack reference with recently active channels

joeyAghion:
* Remove channels that are archived or haven't been active for a while.
* Add a few channels that have been active recently.
* Refresh tips.
* Shorten, slightly.


anandaroop (approved):
👍🏽

PR #547, opened 2025-07-01
docs: Adding #sup-chats to slack.md

chr-tatu:
Adding #sup-chats to slack.md to ensure new starters know about it.


rabidpaul (approved):
Love readmes like this for company docs

PR #581, opened 2026-04-10
docs: remove Peril mentions

amonkhouse:
Peril has been deprecated for a few months now, so I'm cleaning up mentions of it in our docs.


mc-jones (approved):
🙏

PR #582, opened 2026-05-27
RFC: disclose LLM usage in commits or PRs

joeyAghion:
## Summary

Adopt a lightweight convention for marking commits and/or PRs that were produced with significant assistance from an LLM, including which provider and--optionally--model.

## Proposal

Engineers should annotate commits or PRs where an LLM materially contributed to the change, using the [emerging](https://allthingsopen.org/articles/open-source-ai-contributions-assisted-by-git-trailer-standard) `Assisted-by` git trailer convention. Model and version are helpful but optional. E.g.:

```
Assisted-by: Claude:Opus-4.5
```

<details>
  <summary>Original proposal</summary>
Engineers should annotate commits or PRs where an LLM materially contributed to the change, using git's existing `Co-authored-by` pattern. Model and version are helpful but optional. Examples:

```
Co-authored-by: Claude Opus 4.7 <noreply@anthropic.com>
```

```
Co-authored-by: GitHub Copilot <noreply@github.com>
```

```
Co-authored-by: Jane McJaneson <jane@artnet.com>
Co-authored-by: Cursor (claude-sonnet-4-6) <noreply@cursor.sh>
```
</details>

## Reasoning

This creates transparency and maybe even a queryable record of AI involvement in our codebase. LLM-assisted development is already routine, but we have limited visibility into which changes were AI-assisted and which tool or model produced them. Without that signal we cannot:

- Measure the volume and trajectory of AI-assisted contributions across teams and repos
- Correlate AI authorship with downstream signals: release frequency, review latency, defect/rework rates, test coverage changes, incidents, etc.
- Tune our policies (code review, test coverage, security scanning, CI workflows, release gates...) based on real-world experience

## Exceptions

- Not a disclaimer or warning label. AI-assisted code is held to the same standard as any other code. I.e., authors, reviewers, and ultimately teams are fully responsible.
- Not enforced or verified. We don't intend to use this for legal, licensing, or provenance purposes.
- Obviously we're not concerned with past commits or PRs, only future ones.

## Additional Context

~Tools like Claude Code already emit this by default, so the additional lift for users of other tools (or Claude Code users who hand-author commits or PRs) is minimal. Github already renders these coauthors faithfully in its UI and parses them for stats.~ Multiple co-authors are already supported.

Both Artsy and Artnet have adopted conventional commits as a way to structure commit and PR messages. Tools can ostensibly parse this data, and we've analyzed that to understand trends in the past (though the necessary data integrations were disabled during recent migrations). We don't want to constrain engineers, but do want to be able to ask questions about the results in the future. This starts to give us some data to work with.

_Related:_ [EFF's recent policy](https://www.eff.org/about/opportunities/volunteer/coding-with-eff) about disclosing AI involvement in PRs to their open-source projects

## How is this RFC resolved?

I'll leave this open for at least a week for discussion. I'd especially like feedback about:
* Is this too burdensome?
* Are there other formal ways we should consider annotating PRs?

### DIFF
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -81,6 +81,17 @@ Include any details that add context to the PR, as [reviewers likely have less c
 
 If for any reason a PR doesn't contain well-structured commits, [title the PR using the conventional commit format](https://github.com/artsy/README/issues/327) and recommend that it be squashed into `main` rather than merged.
 
+#### Co-authors
+
+It's nice to give credit for pairing by [specifying co-authors](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) in commits and PRs.
+
+When an LLM has materially contributed to a change, our convention is to specify that in an annotation as well. E.g.:
+
+```
+Co-authored-by: Jane McJaneson <jane@artnet.com>
+Co-authored-by: Cursor (claude-sonnet-4-6) <noreply@cursor.sh>
+```
+
 #### Assignees and Reviewers
 
 **Request a review** from at least one team member. This should be someone who has context around the changes that were made, experience with the system(s) being modified, and/or knowledge of the language/framework in use. Feedback is always welcome from team members not specifically requested as well.
--- a/playbooks/engineer-workflow.md
+++ b/playbooks/engineer-workflow.md
@@ -81,15 +81,15 @@ Include any details that add context to the PR, as [reviewers likely have less c
 
 If for any reason a PR doesn't contain well-structured commits, [title the PR using the conventional commit format](https://github.com/artsy/README/issues/327) and recommend that it be squashed into `main` rather than merged.
 
-#### Co-authors
+#### Trailers
 
 It's nice to give credit for pairing by [specifying co-authors](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) in commits and PRs.
 
-When an LLM has materially contributed to a change, our convention is to specify that in an annotation as well. E.g.:
+When an LLM has materially contributed to a change, our convention is to specify that in an `Assisted-by` annotation as well. E.g.:
 
 ```
 Co-authored-by: Jane McJaneson <jane@artnet.com>
-Co-authored-by: Cursor (claude-sonnet-4-6) <noreply@cursor.sh>
+Assisted-by: Claude:Opus-4.5
 ```
 
 #### Assignees and Reviewers


luizonthedrums:
_This comment was assisted by Claude Code_

Maybe it's possible to automate like this by adding on the .claude/settings.json of a repo.

.claude/settings.json

```
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Bash",
        "hooks": [
          {
            "type": "command",
            "command": ".claude/hooks/amend-claude-trailer.sh"
          }
        ]
      }
    ]
  }
}

```
The matcher fires on every Bash tool call; the script itself filters to git-commit invocations so we don't pay any meaningful cost on other shell commands.

.claude/hooks/amend-claude-trailer.sh (chmod +x)

```
#!/bin/bash
set -euo pipefail

payload=$(cat)
command=$(echo "$payload" | jq -r '.tool_input.command // ""')

# Only act on plain `git commit`; skip amends, rebases, merge continuations.
if ! echo "$command" | grep -qE '(^|[[:space:]&;|])git[[:space:]]+commit([[:space:]]|$)'; then
  exit 0
fi
if echo "$command" | grep -qE -- '--amend|--no-edit'; then
  exit 0
fi

# If HEAD doesn't exist (initial commit failed), bail.
msg=$(git log -1 --pretty=%B 2>/dev/null) || exit 0

# Already trailered? Nothing to do.
if echo "$msg" | grep -qiE '^co-authored-by:.*claude'; then
  exit 0
fi

git commit --amend --no-edit \
  --trailer "Co-Authored-By: Claude <noreply@anthropic.com>" >/dev/null

```
A few notes for the PR description when you ship it:

- The --amend after the commit only rewrites SHA before the engineer pushes, so no force-push pain.
- First time an engineer runs Claude Code after this lands, they'll get a one-shot "trust project hooks?" prompt. Worth mentioning in the PR body so it's not a surprise.
- Squash-merge to main/release will still strip the trailer from the resulting commit — that's a GitHub limitation, not something the hook can fix. Your detector already counts at the PR-commit level (it reads commits.message per PR before squash), so this is fine for your metric.

lidimayra:
I like this idea. And from my perspective, as someone who often prefers to write the commit messages myself, I don't think it's burdensome.

I usually rely on https://github.com/jonallured/pear to add collaborators to my messages, so it'd just be a matter of adding it to list (split by model/version).

brainbicycle:
+1, funnily enough i recently stopped doing this after people on the internet started complaining about commit messages looking like advertisements. I don't personally care all that much about that though and think the attribution and signal AI was involved is more important.

mbrowne:
One thing to note is that tools like Cursor have an "auto" option, and if you're using auto it doesn't tell you which LLM model it chose. But personally when I'm using Cursor I usually choose a specific model depending on the use case.

mc-jones:
Make sense to me. Will require a bit of a workflow adjustment, as I also like writting my own commits, even when code is AI generated, but feel it is reasonable. 

@joeyAghion - Not sure if you have an opinion on it, or we should discuss, but what level of granularity are you imagining? Should it be at the commit or PR level? Does it matter?

ftrifoglio:
I've sometimes been so lazy that I let claude commit some changes for me and the co-authored trailer is added automatically. Since then I've revoked permissions on some git commands, but I've tested this git alias and it seems to be working. 

```
git config --global alias.cc '!f() {
  git commit "$@" --trailer "Co-Authored-By: Claude <noreply@anthropic.com>";
}; f'
```

It also works with `git commit --amend`, which is nice if you do `git commit` and want to quickly amend the comment to include Claude.

I personally prefer this approach because it doesn't lock me into or the other. I would just need to remember to use `git commit` or `git cc` depending on whether I want to attribute some authorship to Claude or not.

mbrowne:
Following up on my earlier comment, maybe for auto mode in Cursor, it would make sense to still put something where the name of the model would usually go, and have it say, "Cursor in auto mode". If using Cursor with a specific model like Claude or OpenAI, it probably doesn't need to mention that Cursor was used since it's really the model we would want to track.

BTW, we should also account for any automatic PRs done via Claude or Copilot as part of github code review.

joeyAghion:
> what level of granularity are you imagining? Should it be at the commit or PR level?

Ultimately I'm interested in what lands on `main`, so if a PR's commits are well-structured and can be merged in, I'd want the co-author to be on those commits. If a PR is "squashed," I'm actually not sure if co-authors persist on to the squash commit or now. I'd _hope_ so.

lidimayra:
>  I'm actually not sure if co-authors persist on to the squash commit or now. I'd hope so.

They do. Unless manually changed, the resulting commit of squashed PR keeps 100% of the commit messages within its body.

Recent example from a squashed PR from @olerichter00 here: https://github.com/artsy/volt/commit/feff995c950ad719d9f271ebcf63118b60ec9c0c

dblandin:
I'm in favor of capturing this signal but I have some thoughts on the mechanism.

I actually resonate with what @brainbicycle mentioned: the default `Co-authored-by: Claude` trailer has always read more like an advertisement than a useful signal to me. There's [broader pushback](http://bence.ferdinandy.com/2025/12/29/dont-abuse-co-authored-by-for-marking-ai-assistance/) on this too: `Co-authored-by:` is meant for people, so tagging a tool there overloads the field and still implies authorship it wasn't meant for. Naming the specific model helps, but doesn't fully solve that.

An alternative is `Assisted-by: <model>`, which [a number of larger projects have started using](https://allthingsopen.org/articles/open-source-ai-contributions-assisted-by-git-trailer-standard) (Linux kernel, Fedora, LLVM, OpenTelemetry). It costs slightly more effort since tools often emit `Co-authored-by:` by default, but it keeps the same benefits – still a plain, greppable trailer (`git log --grep="Assisted-by:"`) – and reads as a tool note rather than a co-author. One extended format leaves room for other details: `Assisted-by: AGENT:MODEL [TOOL]`.

Some examples:

```
Assisted-by: Claude:Opus-4.5
Assisted-by: GPT:5.2 [copilot]
Assisted-by: Claude:Sonnet-4.6 [cursor]
```

For automation, Claude Code or other agents could write the current model to a marker (via Stop/PostToolUse hook), and a committed `prepare-commit-msg` hook (shared via husky) appends the trailer at commit time.

Not a blocker on the proposal – mostly flagging that `Assisted-by:` seems to be where the convention is settling, and it avoids the advertisement feel while still giving us a queryable record.

Worth noting there's also an emerging class of tools (e.g. [Git AI](https://github.com/git-ai-project/git-ai)) that attribute AI involvement at the line level rather than per-commit, storing the metadata in git notes so it survives rebases and squashes. That's heavier than what we need for this, but could be worth revisiting as the space matures.

olerichter00:
> An alternative is Assisted-by: <model>, which [a number of larger projects have started using](https://allthingsopen.org/articles/open-source-ai-contributions-assisted-by-git-trailer-standard) (Linux kernel, Fedora, LLVM, OpenTelemetry). It costs slightly more effort since tools often emit Co-authored-by: by default, but it keeps the same benefits – still a plain, greppable trailer (git log --grep="Assisted-by:") – and reads as a tool note rather than a co-author. One extended format leaves room for other details: Assisted-by: AGENT:MODEL [TOOL].

I like the idea of distinguishing between human co-authors and AI assistance. To reduce the effort, we could (and probably should) create a skill with conventions on how to create a PR and adjust the project-based PR templates ([example](https://github.com/artsy/force/blob/main/pull_request_template.md#L1)).

joeyAghion:
I find the case for `Assisted-by` compelling. It avoids needing a list of AI tools in any code seeking to measure "assistance." The drawbacks are:
* Github's UI doesn't recognize that as a contributor. But this is minor and might be temporary. Maybe it's even _better_ that responsibility is visibly and exclusively with the human contributor.
* Tools don't emit this by default. Presumably `AGENTS.md` or other hints can help with this. Any clever parsing and analysis we attempt may have to accept both forms if this continues to be a popular approach.

I'll switch the playbook to recommend `Assisted-by`.

bgastermann:
I generally support this, and the transparency is important.

That being said, I find it a bit annoying to do this on an individual commit-level. Most of the time I let Claude draft a version, I then edit some aspects manually, then commit the change manually. Having to tag each commit that was "inspired by AI" is additional overhead in my perspective. If it's enough to flag this on a PR-level, that would be great.

overengineeredit:
FWIW: I'm supportive for the proposal of adding some attribution/assistance of an LLM.

mzikherman:
I'm for this, but my 2c is that 100% of commits are assisted by at this point, and I'm a tiny bit skeptical of learnings gleaned from that data (although we won't know if we don't have the data - granted).

Have we done any meta/org-level analysis yet on commit frequency, deploy frequency, and seen if there are visible changes in the last 6-9 months? I'm also still a tiny bit 'fuzzy' on what exactly we might _do_ in response, for instance:

>Tune our policies (code review, test coverage, security scanning, CI workflows, release gates...) based on real-world experience

I think it's reasonable to assume that AI is already assisting the vast majority (if not all) PR's and nothing is jumping out at me as immediately 'broken' or 'needs to be updated for the AI age'. It's likely that takes longer to show up (as AI code accumulates, the 'cruftiness' of the codebases grow and eventually velocity will have peaked and declined sort-of thing) - but that's a longer timeline (and we can probably use fall 2025 as the 'tipping point' when we likely moved to majority AI-assisted across the org).

joeyAghion:
> If it's enough to flag this on a PR-level, that would be great.

I think this is OK. PRs are a better unit to measure than commits anyway, so any analysis could inspect the PR description _or_ any individual commits to identify AI assistance.

> Have we done any meta/org-level analysis yet on commit frequency, deploy frequency, and seen if there are visible changes in the last 6-9 months?

I know that @luizonthedrums has started some analysis. The results weren't clear and that partly motivated this RFC.

> ...assume ...likely...

I'd like to get more precise. If the proportion is close to 100% (or close enough that the remainder seems more a tracking error than good data), I'd agree this doesn't add much.

I will merge so we can start to see how much trouble or value this provides. As always, we should reassess when things change.

PR #583, opened 2026-05-28
chore: clarify instructions for adding SSH key in CircleCI

iskounen:
This PR updates the deployments playbook to clarify where to find a private key, because I got confused while following these instructions.

### DIFF
--- a/playbooks/deployments.md
+++ b/playbooks/deployments.md
@@ -78,7 +78,7 @@ README.
     file.
   - Check the _Allow write access_ box and click the _Add key_ button to save the new key.
   - In the CircleCI project settings, go to _Additional SSH Keys_ > _Add SSH Key_.
-  - Enter `github.com` for _Hostname_ and the contents of the _private_ key file for _Private Key_, then click _Add
+  - Enter `github.com` for _Hostname_ and the contents of the _private_ (`.pub` minus the `.pub`) key file for _Private Key_, then click _Add
     SSH Key_ to save.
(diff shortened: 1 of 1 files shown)


narikazu (approved):

playbooks/deployments.md
```diff
@@ -78,7 +78,7 @@ README.
     file.
   - Check the _Allow write access_ box and click the _Add key_ button to save the new key.
   - In the CircleCI project settings, go to _Additional SSH Keys_ > _Add SSH Key_.
-  - Enter `github.com` for _Hostname_ and the contents of the _private_ key file for _Private Key_, then click _Add
+  - Enter `github.com` for _Hostname_ and the contents of the _private_ (`.pub` minus the `.pub`) key file for _Private Key_, then click _Add
```
brainbicycle: find this phrasing a bit confusing tbh what about an example:
(the key that does not end in `.pub`, e.g: `<key_name>.pub 🙅 <keyname> <--- choose this one!`)

brainbicycle: or maybe just (the key that does not end in `.pub`) is enough

ovasdi: I agree, the proposed edit is a bit unclear still.

I like what brian proposed or maybe specify the words _file extension_.

>... and the contents of the _private_ key file (the one without the `.pub` file extension) for _Private Key_

iskounen: Thanks for the advice 👍

PR #584, opened 2026-06-05
RFC: use version constraints for deps

mc-jones:
## Proposal

All repository dependencies should include version constraints.

For example:

```ruby
# Gemfile
gem "elasticsearch", "< 7.14" # https://github.com/elastic/elasticsearch-ruby/issues/1429
gem "estella", "~> 7.0"
gem "omniauth", "1.9.2"
```

```javascript
// package.json
{
	...
	dependencies {
		...
		"dotenv": "4.0.0",
		"easy-peasy": "^6.1.0",
		"ejs": "3.1.10",
		"express": "4.22.1",
	}
}
```

## Reasoning

While Artsy _generally_ does use dependency version constraints (can't speak to Artnet), we have discrepancies between repos and a recent PR generate debate about the necessity of defining version constraints. It would be nice for us to have a position on this for quick decision making.

More generally, dependency management is a constant maintenance task that includes risk when not properly implemented. Untested versions can (and often do) break functionality and/or introduce security vulnerabilities.

Version constraints encourage deliberate decision making around our dependencies and can help prevent unintentional breakages. They also allow us to use native dependency manager functionality, e.g. `bundle update` or `npm update`, which can be helpful for maintenance.

## Exceptions

Internal packages (e.g. `@artsy/palette`) would be exempted, but their own dependencies should include versions.

Code/repo maintainers should ultimately decide if a dependency should be unrestricted, but this should be an exception and its reasoning should be documented with a comment.

This RFC is constrained to `typescript`, `javascript`, `ruby` and `python` dependencies as I can't speak on other language version management.

### Is this really necessary if we're using lockfiles?

- Lockfiles are important for dependency management as they give us deterministic installs, which is critical for our CI/CD pipeline, but they are designed to be modified by the dependency manager the generates them and not modified directly. Not using version constraints can break native dependency manager functionality, e.g. `bundle update`. Manifests are for humans, lockfiles are for computers.

### What is your position on how we should version them? Pin/pessimistic/optimistic versioning?

- This RFC doesn't take a position on which versioning strategy should be used. In practice it will depend on the dependency and any opinions of the code/repo maintainers. It just suggests we should version as the default.

### Has this ever actually been an issue at Artsy?

- I'll admit I was surprised my slack search didn't surface more examples of dependency issues. The last I can recall off the top of my head was the [cython version update fallout](https://artsy.slack.com/archives/CA8SANW3W/p1691420429917559). But I'm guessing this RFC may surface others. If not, I would still argue its good practice as it encourages deliberate decision making around our dependencies.

## How is this RFC resolved?

- If approved, A new github action is added to our repos adding a non-blocking warning for dependency updates that don't use version control.
- If rejected, no action is taken, but this RFC will be preserved for future reference, as needed.

### DIFF
--- a/RFCs/use-version-constraints-for-deps.md
+++ b/RFCs/use-version-constraints-for-deps.md
@@ -42,7 +42,7 @@ Version constraints encourage deliberate decision making around our dependencies
 
 Internal packages (e.g. `@artsy/palette`) would be exempted, but their own dependencies should include versions.
 
-Code/repo maintainers should ultimately decide how if a dependency should have unrestrained version updates, but this should be an exception and its reasoning should be documented.
+Code/repo maintainers should ultimately decide if a dependency should be unrestricted, but this should be an exception and its reasoning should be documented with a comment.
 
 This RFC is constrained to `typescript`, `javascript`, `ruby` and `python` dependencies as I can't speak on other language version management.
 
--- a/RFCs/use-version-constraints-for-deps.md
+++ b/RFCs/use-version-constraints-for-deps.md
@@ -1,0 +1,67 @@
+---
+title: RFC
+description: Use Version Constraints for Dependencies
+---
+
+## Proposal
+
+All repository dependencies should include version constraints.
+
+For example:
+
+```ruby
+# Gemfile
+gem "elasticsearch", "< 7.14" # https://github.com/elastic/elasticsearch-ruby/issues/1429
+gem "estella", "~> 7.0"
+gem "omniauth", "1.9.2"
+```
+
+```typescript
+// package.json
+{
+	...
+	dependencies {
+		...
+		"dotenv": "4.0.0",
+		"easy-peasy": "^6.1.0",
+		"ejs": "3.1.10",
+		"express": "4.22.1",
+	}
+}
+```
+
+## Reasoning
+
+While Artsy _generally_ does use dependency version constraints (can't speak to Artnet), we have discrepancies between repos and a recent PR generate debate about the necessity of defining version constraints. It would be nice for us to have a position on this for quick decision making.
+
+More generally, dependency management is a constant maintenance task that includes risk when not properly implemented. Untested versions can (and often do) break functionality and/or introduce security vulnerabilities.
+
+Version constraints encourage deliberate decision making around our dependencies and can help prevent unintentional breakages. They also allow us to use native dependency manager functionality, e.g. `bundle update` or `npm update`, which can be helpful for maintenance.
+
+## Exceptions
+
+Internal packages (e.g. `@artsy/palette`) would be exempted, but their own dependencies should include versions.
+
+Code/repo maintainers should ultimately decide how if a dependency should have unrestrained version updates, but this should be an exception and its reasoning should be documented.
+
+This RFC is constrained to `typescript`, `javascript`, `ruby` and `python` dependencies as I can't speak on other language version management.
+
+## Additional Context
+
+### Is this really necessary if we're using lockfiles?
+
+- Lockfiles are important for dependency management as they give us deterministic installs, which is critical for our CI/CD pipeline, but they are designed to be modified by the dependency manager the generates them and not modified directly. Not using version constraints can break native dependency manager functionality, e.g. `bundle update`. Manifests are for humans, lockfiles are for computers.
+
+### What is your position on how we should version them? Pin/pessimistic/optimistic versioning?
+
+- This RFC doesn't take a position on which versioning strategy should be used. In practice it will depend on the dependency and any opinions of the code/repo maintainers. It just suggests we should version as the default.
+
+### Has this ever actually been an issue at Artsy?
+
+- I'll admit I was surprised my slack search didn't surface more examples of dependency issues. The last I can recall off the top of my head was the [cython version update fallout](https://artsy.slack.com/archives/CA8SANW3W/p1691420429917559). But I'm guessing this RFC may surface others. If not, I would still argue its good practice as it encourages deliberate decision making around our dependencies.
+
+## How is this RFC resolved?
+
+- If approved, A new github action is added to our repos adding a non-blocking warning for dependency updates that don't use version control.
+- If rejected, no action is taken, but this RFC will be preserved for future reference, as needed.
+


bgastermann:
I fully agree and support this. 👍

As a side-note: this is not an issue for .NET/C# (Artnet's back-end implementations) as packages there are pinned to a specific version by default until explicitly updated. I can't comment on Artnet's front-end, though, and leave it up to @mbrowne and/or @mjtedder to speak to the approach used there.

leamotta:
I shared my perspective in detail in the [original PR discussion](https://github.com/artsy/gravity/pull/20171#issuecomment-4555195684) about removing a constraint from the Gemfile, that thread is still relevant and gives more background to my thoughts. Addressing the RFC specifically, and focusing on the impact for Rails projects:

#### On "manifests are for humans, lockfiles are for computers"

I agree, and they should carry information meaningful to humans: _why_ a constraint exists. A bare `~> 5.0` tells a future maintainer very little. A comment like `# held below 6 due to redis 6 API incompatibility` tells them exactly when it's safe to revisit.

#### On "Version constraints encourage deliberate decision making and can help prevent unintentional breakages"

This is useful on scenarios where the constraint is there for a proven reason, yet the RFC proposes to do this for every dependency (including deps that are installed at their latest available version). The RFC doesn't explain the mechanism by which removing a version constraint in this latter scenario causes an accidental bump when the lockfile already pins the exact version and any update requires merging a dependabot PR or running an explicit `bundle update <gem>`, both going through code review. Also not sure what is meant by _"not using version constraints can break native dependency manager functionality e.g. `bundle update`"_  perhaps an example can help clarify?

#### On the practical impact

This is where I have the strongest feelings. If approved, this RFC would require repos like [Exchange](https://github.com/artsy/exchange/blob/main/Gemfile), which have been carefully maintained and have targeted pins where they matter, to add version constraints to every dependency, plus adopt a new GitHub Action to enforce it. That's a non-trivial maintenance burden for protection against a scenario the RFC acknowledges it couldn't find examples of. The lockfile already provides deterministic installs; manifest constraints would be redundant while adding ongoing overhead every time a dependency is intentionally upgraded.

Out of curiosity, I looked at some well-known open source Rails projects, they seem to have a pretty lax approach. Why do we need to be stricter?

- https://github.com/discourse/discourse/blob/main/Gemfile
- https://github.com/mastodon/mastodon/blob/main/Gemfile
- https://github.com/basecamp/fizzy/blob/main/Gemfile

#### What I'd support

Version constraints when there's a known compatibility boundary, with a comment explaining why. That produces Gemfiles that are genuinely useful to future maintainers. We've been doing this informally in Rails repos and I don't think it requires the overhead of a GitHub Action on top yet, but open for discussion.

mc-jones:
>The RFC doesn't explain the mechanism by which removing a version constraint in this latter scenario causes an accidental bump when the lockfile already pins the exact version and any update requires merging a dependabot PR or running an explicit bundle update <gem>, both going through code review.

Here is scenario I've run into when performing bulk security updates: 

There is a security vulnerability for gem `Foo@1.2.3`. It has been patched in all major versions  but `Foo@latest` is `3.0.1`. If I run `bundle update foo` `bundler` will pull in `Foo@3.0.1`. We've now unnecessarily updated our repo to two major versions. This increases the probability of introducing a breaking change and follow up work to diagnose the issue. If it was pessimistically version constrained, it would have just pulled in the  security patch (e.g. `Foo@1.2.4`). 

This would also help us better automate dependency updates. 

Re code review, it is my experience that reviewers leave it to the submitter to test dependency compatibility and assume there is a reason for the version being chosen, so I believe reviews are being over-indexed. 

There is also a more theoretical argument being made here that we should aim to be deliberate in our dependency selection. Adding version constraints is a self documenting way for a engineer to say, I tested and have a reason for including this dependency at this version. That gets lost with full reliance on lock files. 

> Also not sure what is meant by "not using version constraints can break native dependency manager functionality e.g. bundle update" perhaps an example can help clarify?

Without version constraints every single gem gets bumped to `latest`. It is rare for this to ever be intended, which in practice makes the operation useless. 

>That's a non-trivial maintenance burden for protection against a scenario the RFC acknowledges it couldn't find examples of

This feels like a uncharitable reading of what was said. There is an example included and I was aiming to not overstate the issue. I don't think everything is captured in slack, which was more the point.

> Out of curiosity, I looked at some well-known open source Rails projects, they seem to have a pretty lax approach. Why do we need to be stricter?

I could flip this argument as each example does include version constraints. In particular, the vast majority of [mastadon](https://github.com/mastodon/mastodon/blob/main/Gemfile) gems. But I can agree that the value of this approach can vary by language and their dependency resolution handling. Again, the RFC isn't meant to be all or nothing but, rather, a default that can be deviated from depending on need.

mbrowne:
For JS, I find it helpful to make use of available syntax for specific version constraints, i.e. `^` and `~` to indicate major and minor versions, or just pinning a specific version if there's a reason a specific patch version needs to be used. I think this is especially relevant with JS projects because comments are invalid in JSON files. The lock file alone only tells you what version was installed the last time someone happened to update a particular package, but it doesn't communicate things like, "We had to install a slightly earlier version of this package for compatibility reasons".

(Of course, there's only so much you can communicate just with a version constraint, so for things that might be tricky or confusing, devs should also make a note in the readme.)

jpotts244:
I agree with @leamotta's callouts. For Ruby/Rails applications that use Bundler, this seems to move away from the dependency management model the ecosystem is designed around.

**Bundler's model is:**

```text
Gemfile      -> acceptable versions
Gemfile.lock -> exact versions
```

If we pin every gem:

```text
Gemfile      -> exact versions
Gemfile.lock -> exact versions
```

then both files are serving nearly the same purpose.

In my experience, exact version pins in the `Gemfile` are typically reserved for special cases (e.g. known compatibility issues, regressions, or temporary constraints during an upgrade). In those situations, the pin also helps tell the story of why that dependency is being constrained.

Since `Gemfile.lock` already provides exact-version reproducibility, can we clarify what additional guarantee we gain from pinning every gem in the `Gemfile`? My concern is that we end up duplicating version management while increasing the maintenance cost of dependency upgrades and introducing additional churn when updating gems.

mc-jones:
@jpotts244 I just want to clarify that the proposal is not to **pin** versions. The proposal is that we default to including a version constraint, which includes ranges, e.g. `"~> 7.0"`. In this way it aligns with the bundler dependency management you describe:

```
Bundler's model is:

Gemfile      -> acceptable versions 
Gemfile.lock -> exact versions
```

Additionally, it doesn't suggest we never use unrestricted versioning (i.e. Not including any version constraint).

narikazu:
A few thoughts to add to this discussion:

`bundle update --patch` and `bundle update --minor` already exist for limiting update scope, so the concern about unintended major version bumps can be addressed at the command level without requiring manifest constraints.

If the RFC's main motivation is preventing accidental major upgrades via `bundle update`, then any constraint that would actually achieve this needs to include an upper bound (e.g. `~> x.y`, `< 2.0`). The RFC says it takes no position on versioning strategy, which means a bare `>= 0` or lower-bound-only constraint would technically satisfy the proposal while doing nothing to prevent the scenario it describes. It might be worth clarifying what kind of constraint actually achieves the stated goal. 🤔 

On the Mastodon comparison: having vast majority of gems constrained is a bit different from requiring _all gems to have constraints_. Looking at major OSS Rails projects Leandro showed, none of them seem to meet the bar proposed here.

It's also not entirely clear what the benefit would be of applying constraints to the dependencies in the `development` or `test` group that are not installed to staging and production environments.

More broadly: for a change that adds non-trivial maintenance cost across all repos, it would help to see at least one case where the absence of a version constraint led to an unintended upgrade reaching production, something that `bundle update`, Dependabot, code review, and staging deployment all failed to catch. Without a concrete example, it's hard to assess the actual risk being addressed.

mc-jones:
Seems clear that I'm in the minority here. Closing.

I'll disagree and commit.

PR #589, opened 2026-08-31
RFC: Track where 1Password secrets/tokens are used

ovasdi:
## Summary
- Adds an RFC proposing that every 1Password item for a shared secret/token include a "Used in" field (every consumer) and a "Source" field (where it's generated/managed, which account owns it), plus a short rotation checklist.
- Full draft with examples: https://app.notion.com/p/3c9cab0764a08166893bc14fcbd67527
- Background discussion: see Slack threads linked in the RFC's Additional Context section.

Per the [RFC process](https://github.com/artsy/README/blob/main/playbooks/rfcs.md), this will be announced at the next dev standup and left open for at least a week for feedback.

Assisted-by: Claude:Sonnet-5 [Claude Code]

### DIFF
--- a/RFCs/1password-secret-source.md
+++ b/RFCs/1password-secret-source.md
@@ -1,0 +1,40 @@
+---
+title: RFC
+description: Track where 1Password secrets/tokens are used
+---
+
+## Proposal
+
+Every 1Password item for a shared secret/token should include two fields: "Used in" (every place the secret is
+consumed) and "Source" (where it's generated/managed and which account owns it), plus a short checklist for
+rotating it. Full draft with examples: [1Password Secret & Token Hygiene](https://app.notion.com/p/3c9cab0764a08166893bc14fcbd67527)
+
+## Reasoning
+
+When we need to rotate a secret stored in 1Password, it's often unclear where else that secret is set — CI env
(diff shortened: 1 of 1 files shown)


MounirDhahri:
Thanks for opening this RFC. Having gone through the trouble of figuring out where a key is used across multiple systems is painful. This is a welcome addition for me.

paulness:
I agree with this change, great suggestion. @ovasdi

paulness (approved):

bgastermann:
I wholeheartedly agree with this; should make our lives easier, though I'm afraid people will forget to set this in the beginning. Is there some way to "enforce" or check for/remind about this?

UPDATE:
I left a comment on the Notion page but after checking the README ([exceptions](https://github.com/artsy/README/pull/589/changes#diff-0e9232ad4c2ecc68a3e3771312fbd2c2350f65a03a0e63c3c09ccf167e77801dR21) section) again I think all of my questions got answered.

bgastermann (approved):

ovasdi:
@bgastermann 

> I'm afraid people will forget to set this in the beginning

Totally valid concern and I have it too. But this is a step in a direction that hopefully with time becomes a common practice even if initially it isn't followed through 100% of the time. 😄

PR #590, opened 2026-08-31
RFC: Deprecate completed_onboarding_at

iskounen:
Proposes removing `completed_onboarding_at` from gravity's `User` model (and the corresponding metaphysics mutation input, and every place eigen/force set it).

- The field was added for [GRO-125](https://artsyproduct.atlassian.net/browse/GRO-125) to let **Sailthru** know onboarding status.
- The only current consumer is a sync to **Braze**, and I've been told Marketing isn't actually using this attribute there.
- We already have analytics events that give a much richer picture (start / step completion / skip / abandon) than a single timestamp.
- The flag isn't even consistent across platforms: on the app it's really just "how long before this forced flow was exited," and on web it's "reached a terminal screen, sometimes." Neither reliably means "completed onboarding."

Full reasoning and next steps are in the RFC doc itself.

Assisted-by: Claude:Sonnet-5

[GRO-125]: https://artsyproduct.atlassian.net/browse/GRO-125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RFCs/deprecate-completed-onboarding-at.md
+++ b/RFCs/deprecate-completed-onboarding-at.md
@@ -1,0 +1,69 @@
+---
+title: Deprecate completed_onboarding_at
+description: Removing the completed_onboarding_at flag on gravity's User model, since it no longer tells us anything reliable.
+---
+
+## Proposal
+
+I want to deprecate and remove `completed_onboarding_at` on gravity's `User` model, along with the
+`completedOnboarding` mutation input on metaphysics and every place in eigen and force that sets it.
+
+## Reasoning
+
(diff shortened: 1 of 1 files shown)


anandaroop:
You had me at: _the timestamp doesn't tell you whether someone "completed" anything meaningful_

mbrowne:
Makes sense not to keep a misleading/inaccurate flag—LGTM.

iskounen:
I've confirmed that `completed_onboarding_at` is not used in any Braze campaigns, canvases or segments. This seems like more of a JFDI task than an RFC, so closing.

